PR #46, opened 2014-12-17
Replace AbstractMongoTest with JUnit Externalresource

dcrissman:
I went to use test. AbstractMongoTest and realized that it was broken and nothing actually used it anyway. So I would like to replace it with a JUnit ExternalResource.

I also noticed that there are several versions of AbstractMongoTest in different places that are not all the same. If this change is accepted, I will make a story to replace all the copy/paste versions with the new ExternalResource.


dcrissman:
I believe the open concerns have been addressed at this point. Unless there is anything else, I would like to re-submit for review.

PR #56, opened 2015-01-15
Fix #55: Save qps as string

### LINKED ISSUE #55: Logical operators in queries in metadata cannot be saved to MongoDB
Metadata with query:[ "$and": [ ...] } causes this:

Document field names can't start with '$' (Bad Key: '$and')"

We have to save query/projection/sort as string.

bserdar:


jewzaam:
Isn't there still a problem with saving?  Thought `putProjection` would need updated as well to save the projection, sort, and query as string instead of bson.

jewzaam:
Would be good to have a unit test added for the case that query (part of association) contains a key with `$` in it and try persisting the metadata.

bserdar:
You are correct. Fixed now.

bserdar:
Unit test is also added.

metadata/src/test/java/com/redhat/lightblue/metadata/mongo/MongoMetadataTest.java
```diff
@@ -173,6 +173,25 @@ public void createMdTest() throws Exception {
...
 
+    @Test
+    public void createMdWithAndRefTest() throws Exception {
+        Extensions<JsonNode> extensions = new Extensions<>();
+        extensions.addDefaultExtensions();
+        extensions.registerDataStoreParser("mongo", new MongoDataStoreParser<JsonNode>());
+        JSONMetadataParser parser = new JSONMetadataParser(extensions, new DefaultTypes(), new JsonNodeFactory(true));
+
+        // get JsonNode representing metadata
+        JsonNode jsonMetadata = AbstractJsonNodeTest.loadJsonNode(getClass().getSimpleName() + "-qps-andquery.json");
```
jewzaam: Test file missing.  I'll merge this anyway with this test ignored so builds are not broken.  It's blocking work @derek63 is doing for hystrix.

jewzaam: Issue to fix the test added: https://github.com/lightblue-platform/lightblue-mongo/issues/57

PR #59, opened 2015-01-16
Fixes #28: updated artifactId's

### LINKED ISSUE #28: Add flag to indexes in metadata to indicate if index exists in the backend
Seeing a case where indexes in entityInfo exist that are not in the collection.  Propose either include a field in the metadata when requested from lightblue to indicate if the index actually exists or have a way to fix the index.  A flag is easier...

jewzaam:


jewzaam:
Requires https://github.com/lightblue-platform/lightblue-core/pull/290

luan-cestari:
Note: Travis build failed (even after the update of the core PR) but suddenly (after asking to build again) it worked. The can be something related to the other issue we get often in the rest repository (that travis fails many times to build it correctly).

PR #65, opened 2015-01-20
Fixes #29: delete indexes in mongo no longer referenced by metadata

### LINKED ISSUE #29: Removing index from metadata doesn't remove it from collection
With the exception of _id index my expectation is that if I remove an index from metadata it would be removed from mongo.

jewzaam:
Will be another PR for lightblue-rest for enabling integration tests for index deletion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ util/nb-configuration.xml
 nb-configuration.xml
 metadata/nb-configuration.xml
 docs/cicd/*.png
+*versionsBackup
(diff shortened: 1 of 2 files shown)


bserdar:
A simpler way to deal with ID index would be this, instead of the big IF statement at the end:

 boolean createIx = !isIdIndex(index);
if(!createIx) deleteIndexes.remove(existingIndex)

jewzaam:
`isIdIndex` operates on the lightblue `Index` object.  In the case of deleting an index we do not have that object in the `entityInfo`.

PR #67, opened 2015-01-22
Array field comparison

bserdar:
Requires core #297

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
@@ -734,7 +734,7 @@ public class Translator {
         "for(var i=0;i<n;i++) {"+
         "  if(this.f1[i] >= this.f2[i]) { allOk=false;break;}"+
         "}"+
-        "if(allOk&&(this.%s.length <= this.%s.length)) return true;";
+        "if(allOk&&(this.f1.length <= this.f2.length)) return true;";
 
     private static final String ARR_ARR_GT="var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
         "var allOk=true;"+
@@ -777,7 +777,7 @@ public class Translator {
     }
 
     private String writeArrayFieldComparisonJS(String field,String array,String op) {
-        return String.format("for(var i=0;i<this.%s.length;i++) { if(!(this.%s %s this.%s[i])) return false; } return true;}",array,field,op,array);
+        return String.format("for(var i=0;i<this.%s.length;i++) { if(!(this.%s %s this.%s[i])) return false; } return true;",array,field,op,array);
     }
 
     private String writeComparisonJS(Path field1,boolean field1IsArray,
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
@@ -777,7 +777,7 @@ public class Translator {
     }
 
     private String writeArrayFieldComparisonJS(String field,String array,String op) {
-        return String.format("for(var i=0;i<this.%s.length;i++) { if(this.%s %s this.%s[i]) return true;}",array,field,op,array);
+        return String.format("for(var i=0;i<this.%s.length;i++) { if(!(this.%s %s this.%s[i])) {return false;} else {return true;}}",array,field,op,array);
     }
 
     private String writeComparisonJS(Path field1,boolean field1IsArray,
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
@@ -713,71 +713,65 @@ public class Translator {
         return arr.toString();
     }
 
-    private static final String ARR_ARR_EQ=
-        "if(this.%s.length==this.%s.length) { "+
+    private static final String ARR_ARR_EQ= "if(this.f1.length==this.f2.length) { "+
         "  var allEq=true;"+
-        "  for(var i=0i<this.%s.length;i++) { "+
-        "     if(this.%s[i] != this.%s[i]) { allEq=false; break; } "+
+        "  for(var i=0;i<this.f1.length;i++) { "+
+        "     if(this.f1[i] != this.f2[i]) { allEq=false; break; } "+
         "  } "+
         "  if(allEq) return true;"+
         "}";
 
-    private static final String ARR_ARR_NEQ=
-        "if(this.%s.length==this.%s.length) { "+
+    private static final String ARR_ARR_NEQ="if(this.f1.length==this.f2.length) { "+
         "  var allEq=true;"+
-        "  for(var i=0i<this.%s.length;i++) { "+
-        "     if(this.%s[i] != this.%s[i]) { allEq=false; break; } "+
+        "  for(var i=0;i<this.f1.length;i++) { "+
+        "     if(this.f1[i] != this.f2[i]) { allEq=false; break; } "+
         "  } "+
         "  if(!allEq) return true;"+
         "} else { return true; }";
 
-    private static final String ARR_ARR_LT=
-        "var n=this.%s.length<this.%s.length?this.%s.length:this.%s.length;"+
+    private static final String ARR_ARR_LT="var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
         "var allOk=true;"+
         "for(var i=0;i<n;i++) {"+
-        "  if(this.%s[i] >= this.%s[i]) { allOk=false;break;}"+
+        "  if(this.f1[i] >= this.f2[i]) { allOk=false;break;}"+
         "}"+
-        "if(allOk&&(this.%s.length <= this.%s.length) return true;";
+        "if(allOk&&(this.%s.length <= this.%s.length)) return true;";
 
-    private static final String ARR_ARR_GT=
-        "var n=this.%s.length<this.%s.length?this.%s.length:this.%s.length;"+
+    private static final String ARR_ARR_GT="var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
         "var allOk=true;"+
         "for(var i=0;i<n;i++) {"+
-        "  if(this.%s[i] <= this.%s[i]) { allOk=false;break;}"+
+        "  if(this.f1[i] <= this.f2[i]) { allOk=false;break;}"+
         "}"+
-        "if(allOk&&(this.%s.length >= this.%s.length) return true;";
+        "if(allOk&&(this.f1.length >= this.f2.length)) return true;";
 
-    private static final String ARR_ARR_LTE=
-        "var n=this.%s.length<this.%s.length?this.%s.length:this.%s.length;"+
+    private static final String ARR_ARR_LTE="var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
         "var allOk=true;"+
         "for(var i=0;i<n;i++) {"+
-        "  if(this.%s[i] > this.%s[i]) { allOk=false;break;}"+
+        "  if(this.f1[i] > this.f2[i]) { allOk=false;break;}"+
         "}"+
-        "if(allOk&&(this.%s.length <= this.%s.length) return true;";
+        "if(allOk&&(this.f1.length <= this.f2.length)) return true;";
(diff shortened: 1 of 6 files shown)


jewzaam:
@bserdar there are no unit tests added for this.  Given how it's implemented, with javascript passed to mongo, it would be nice to see that it works.

Also, it there no better way to build the scripts?  This looks like it works but it's not easy to follow.  There is no field binding the formatted strings, which makes this feel brittle long term.  How about a string with escaped bind parameters and use a string replace to set them when used?

```java
    private static final String ARR_ARR_EQ=
        "if(this.:field1:.length==this.:field2:.length) { "+
        "  var allEq=true;"+
        "  for(var i=0i<this.:field1:.length;i++) { "+
        "     if(this.:field1:[i] != this.:field2:[i]) { allEq=false; break; } "+
        "  } "+
        "  if(allEq) return true;"+
        "}";
...
    private String writeArrayArrayComparisonJS(String field1,String field2,BinaryComparisonOperator op) {
        String script = null;
        switch(op) {
        case _eq:
            script = ARR_ARR_EQ;
            break;
        case _neq:
            script = ARR_ARR_NEQ;
            break;
        case _lt:
            script = ARR_ARR_LT;
            break;
        case _gt:
            script = ARR_ARR_GT;
            break;
        case _lte:
            script = ARR_ARR_LTE;
            break;
        case _gte:
            script = ARR_ARR_GTE;
            break;
        default:
            return null;
        }
        return script.replaceAll(":field1:", field1).replaceAll(":field2:", field2);
    }
```

bserdar:
I'll write a test for it,

Bind params as above looks inefficient, with multiple string searches,
maybe I can find a better way of implementing this and make it more
readable.

jewzaam:
I tested it:
```java
    /**
     * Wanted to know how fast two String replaceAll's were compared to
     * String.format with 8 params, re:
     * https://github.com/lightblue-platform/lightblue-mongo/pull/67#issuecomment-73765251
     */
    @Test
    public void format_vs_replaceAll() {
        //return String.format(ARR_ARR_GTE,,,,,,,,);
        String FORMAT = "var n=this.%s.length<this.%s.length?this.%s.length:this.%s.length;"
                + "var allOk=true;"
                + "for(var i=0;i<n;i++) {"
                + "  if(this.%s[i] < this.%s[i]) { allOk=false;break;}"
                + "}"
                + "if(allOk&&(this.%s.length >= this.%s.length) return true;";

        String REPLACE = "var n=this.:field1:.length<this.:field2:.length?this.:field1:.length:this.:field2:.length;"
                + "var allOk=true;"
                + "for(var i=0;i<n;i++) {"
                + "  if(this.:field1:[i] < this.:field2:[i]) { allOk=false;break;}"
                + "}"
                + "if(allOk&&(this.:field1:.length >= this.:field2:.length) return true;";

        int iterations = 1000000;
        String field1 = "f1";
        String field2 = "f2";

        {
            long startTime = System.currentTimeMillis();
            for (int i = 0; i < iterations; i++) {
                String x = String.format(FORMAT, field1, field2, field1, field2, field1, field2, field1, field2);
            }
            long endTime = System.currentTimeMillis();
            System.out.println("FORMAT: " + (endTime - startTime));
        }

        {
            long startTime = System.currentTimeMillis();
            for (int i = 0; i < iterations; i++) {
                String x = REPLACE.replaceAll(":field1:", field1).replaceAll(":field2:", field2);
            }
            long endTime = System.currentTimeMillis();
            System.out.println("REPLACE: " + (endTime - startTime));
        }
    }
```

Output shows the replace is almost twice as fast over 1,000,000 iterations:
```
FORMAT: 4502
REPLACE: 2544
```

bserdar:
allright, we'll replace then.

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -693,45 +713,154 @@ private String writeJSForLoop(StringBuilder bld, Path p, String varPrefix) {
         return arr.toString();
     }
 
-    private DBObject translateFieldComparison(FieldComparisonExpression expr) {
-        StringBuilder str = new StringBuilder(128);
+    private static final String ARR_ARR_EQ=
+        "if(this.%s.length==this.%s.length) { "+
+        "  var allEq=true;"+
+        "  for(var i=0i<this.%s.length;i++) { "+
```
jewzaam: Missing semicolon:  for(var i=0 **;** i<this.%s.length;i++)

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -693,45 +713,154 @@ private String writeJSForLoop(StringBuilder bld, Path p, String varPrefix) {
...
+        "     if(this.%s[i] != this.%s[i]) { allEq=false; break; } "+
+        "  } "+
+        "  if(allEq) return true;"+
+        "}";
+
+    private static final String ARR_ARR_NEQ=
+        "if(this.%s.length==this.%s.length) { "+
+        "  var allEq=true;"+
+        "  for(var i=0i<this.%s.length;i++) { "+
```
jewzaam: Missing semicolon:  for(var i=0 **;** i<this.%s.length;i++)

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -693,45 +713,154 @@ private String writeJSForLoop(StringBuilder bld, Path p, String varPrefix) {
...
+        "     if(this.%s[i] != this.%s[i]) { allEq=false; break; } "+
+        "  } "+
+        "  if(!allEq) return true;"+
+        "} else { return true; }";
+
+    private static final String ARR_ARR_LT=
```
jewzaam: One missing paren in the end if, but could be optimized by pulling the condition up.  Likely similar for GT, LTE, and GTE.

```
    "if(this.%s.length <= this.%s.length) {"+
        "var n=this.%s.length<this.%s.length?this.%s.length:this.%s.length;"+
        "var allOk=true;"+
        "for(var i=0;i<n;i++) {"+
        "  if(this.%s[i] >= this.%s[i]) { allOk=false;break;}"+
        "}"+
        "if(allOk) return true;"+
    "}";
```

jewzaam:
Was a bit strange how array comparison is done, so write it down here: 
https://github.com/lightblue-platform/lightblue-docs-specs/pull/1

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -693,45 +713,154 @@ private String writeJSForLoop(StringBuilder bld, Path p, String varPrefix) {
...
+        case _lte:
+            return String.format(ARR_ARR_LTE,field1,field2,field1,field2,field1,field2,field1,field2);
+        case _gte:
+            return String.format(ARR_ARR_GTE,field1,field2,field1,field2,field1,field2,field1,field2);
+        }
+        return null;
+    }
+
+    private String writeArrayFieldComparisonJS(String field,String array,String op) {
+        return String.format("for(var i=0;i<this.%s.length;i++) { if(this.%s %s this.%s[i]) return true;}",array,field,op,array);
```
jewzaam: I think this needs some testing.  Is the intent that if any element in the array satisfies "field op array[i]" that it's a match?

bserdar: All elements of the array should satisfy the relation:

result is true if for all i, field op array[i]

bserdar: I fixed the JS errors. I'll write tests for it later.

jewzaam: `return String.format("for(var i=0;i<this.%s.length;i++) { if(this.%s %s this.%s[i]) return true;}",array,field,op,array);`

How does this result in "true if for all i, field op array[i]"?  It returns true when it finds one case of field op array[i].

bserdar: It doesn't. Fix pushed.

jewzaam:
For constants ARR_ARR_LT, ARR_ARR_GT, ARR_ARR_LTE, and ARR_ARR_GTE why does the final if contain a check for array length?  That check could be done earlier.  If it isn't true, there's no point doing the other work.  For example, this:

```
    private static final String ARR_ARR_LT=
        "var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
        "var allOk=true;"+
        "for(var i=0;i<n;i++) {"+
        "  if(this.f1[i] >= this.f2[i]) { allOk=false;break;}"+
        "}"+
        "if(allOk&&(this.f1.length <= this.f2.length)) return true;";
```

could become this:
```
    private static final String ARR_ARR_LT=
        "if (this.f1.length <= this.f2.length) {" +
        "  var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
        "  var allOk=true;"+
        "  for(var i=0;i<n;i++) {"+
        "    if(this.f1[i] >= this.f2[i]) { allOk=false;break;}"+
        "  }"+
        "  if(allOk) return true;" +
        "}";
```

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -693,45 +713,148 @@ private String writeJSForLoop(StringBuilder bld, Path p, String varPrefix) {
...
+        "  } "+
+        "  if(!allEq) return true;"+
+        "} else { return true; }";
+
+    private static final String ARR_ARR_LT="var n=this.f1.length<this.f2.length?this.f1.length:this.f2.length;"+
+        "var allOk=true;"+
+        "for(var i=0;i<n;i++) {"+
+        "  if(this.f1[i] >= this.f2[i]) { allOk=false;break;}"+
+        "}"+
+        "if(allOk&&(this.%s.length <= this.%s.length)) return true;";
```
jewzaam: Missed some "%s" in ARR_ARR_LT

bserdar: Fixed that, and started adding some unit test. Re: array length comparison,
i'll look at that

bserdar:
The array length check should be at the end. The for-loop iterates
min(length1,length2) times. If 'allOk' is true, that many elements pass the
test. Only if 'allOk' is true the array length comparison is required,
because then the longer array is "larger".

bserdar:
More unit tests are pushed.

Maybe we should rethink how inequalities are evaluated for arrays. With the
current definitions, it is possible that:

a<b:false and a=b:false, but a<=b:true

This happens if a is a proper prefix of b. a<b fails, because all elements
of a are already in b. a=b fails because array sizes are different. But,
a<=b is true, because a[i]<=b[i] for all i, and a.length<b.length.

jewzaam:
Regarding array length check, I don't understand why it should be done at the end.  If the condition is not met it doesn't matter how many times the for loop iterates, the block will not evaluate to true.  Therefore, why spend the time to loop when it won't matter?

jewzaam:
Regarding inequalities, that's kind of why I created http://jewzaam.gitbooks.io/lightblue-specifications/content/language_specification/field_comparisons_with_arrays.html because it seemed a bit strange.  Maybe that doc should be updated first to reflect what is desired then implement it here?

bserdar:
How about this:

For all comparisons other than !=
     a op b is true if a.length=b.length and a[i] op b[i] for all i
For !=
    a != b is true if either a.length!=b.length or there exists i with
a[i]!=b[i]

jewzaam:
I think lt[e] and gt[e] operations are a bit strange for comparing arrays when it gets to the content.  Should we support them at all?  I don't know that they're useful.  Set operations though would be helpful, but I wouldn't expect that as part of this PR.  Given questionable use of lt[e] and gt[e] I'd say what you propose is fine, the key point being it's very clearly documented so there's no misunderstanding.

bserdar:
allright, i'll fix the doc and the code.

bserdar:
I unintentionally pushed the doc fixes to master branch of the upstream
repo, please review them. Sorry.

bserdar:
The PR for mongo is updated, there is a new PR for core.

bserdar:
Oh wait, I didn't push to master...

jewzaam:
@bserdar missing resource `testMetadata5.json`

jewzaam:
Also missing `testdata5.json`

PR #78, opened 2015-03-04
Fix hystrix java maven problem

luan-cestari:
Related to the issue https://github.com/lightblue-platform/lightblue-core/pull/336 , which in summary we had to upgrade the hystrix version which cause a little change on its interface


luan-cestari:
restarted as the PR on core was merged

jewzaam:
We added the override you're removing because we wanted to get at the underlying exception to throw the cause, as shown.  Do we have tests that show how code behaves when `HystrixBadRequestException` is thrown instead of `RuntimeException`?

luan-cestari:
@jewzaam yeah, I notice the comment but I think I read something that Hystrix change a bit on the exception handling (not sure if it is the reason we didn't have any test broken when I amde the change, do you know @bserdar  if we have those tests?) but the point of this change is that method is final, so we can't override it anymore

PR #81, opened 2015-03-10
Rest issues13incorrect error code

luan-cestari:
To solve the problem with authentication with the remote/local database described in https://github.com/lightblue-platform/lightblue-rest/issues/13 , create this PR which it depends on https://github.com/lightblue-platform/lightblue-core/pull/339 (a change on the core repository). Spend some time making the test for this change but found a blocker that  de.flapdoodle.embed.mongo doesn't allow authentication, so I created an issue (  https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/issues/114 ) and left the draft for the test (which doesn't impact the other tests) to be resumed as soon as the dependency fix that problem and we update to the necessary version to have that feature.

### DIFF
--- a/metadata/pom.xml
+++ b/metadata/pom.xml
@@ -31,6 +31,12 @@
     <name>lightblue-mongo: ${project.groupId}|${project.artifactId}</name>
     <dependencies>
         <dependency>
+            <groupId>com.binarytweed</groupId>
+            <artifactId>quarantining-test-runner</artifactId>
+            <version>0.0.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>com.redhat.lightblue.mongo</groupId>
             <artifactId>lightblue-mongo-common</artifactId>
             <version>${project.version}</version>
--- a/metadata/pom.xml
+++ b/metadata/pom.xml
@@ -33,7 +33,6 @@
         <dependency>
             <groupId>com.binarytweed</groupId>
             <artifactId>quarantining-test-runner</artifactId>
-            <version>0.0.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


jewzaam:
Rebuilding

metadata/pom.xml
```diff
@@ -31,6 +31,12 @@
     <name>lightblue-mongo: ${project.groupId}|${project.artifactId}</name>
     <dependencies>
         <dependency>
+            <groupId>com.binarytweed</groupId>
+            <artifactId>quarantining-test-runner</artifactId>
+            <version>0.0.1</version>
```
jewzaam: Put quarantining-test-runner dependency in root pom's `dependencyManagement` and remove version from here.

luan-cestari:
Changed

metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
```diff
@@ -223,6 +216,15 @@ public EntityInfo getEntityInfo(String entityName) {
         } catch (Error e) {
             // rethrow lightblue error
             throw e;
+
+        } catch (CommandFailureException e) {
+            // give a better exception in case auth failed
+            LOGGER.error(e.getMessage(), e);
+            if(e.getCode() == 18){
```
jewzaam: Where did you get a code of 18?

luan-cestari: It is from mongodb, https://github.com/lightblue-platform/lightblue-rest/issues/13 and we can also simulate it, but it is the MongoDB itself that generate this message for this case https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err#L21

PR #84, opened 2015-03-13
Error code timeoutrest#12

luan-cestari:
This PR is to fix the issue https://github.com/lightblue-platform/lightblue-rest/issues/12 and also it enhanced what I already did on the issue https://github.com/lightblue-platform/lightblue-rest/issues/13

### DIFF
--- a/metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
+++ b/metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
@@ -18,75 +18,40 @@
  */
 package com.redhat.lightblue.metadata.mongo;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.HashSet;
-import java.util.regex.Pattern;
-
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.mongodb.*;
+import com.redhat.lightblue.DataError;
+import com.redhat.lightblue.OperationStatus;
+import com.redhat.lightblue.Response;
+import com.redhat.lightblue.common.mongo.MongoDataStore;
+import com.redhat.lightblue.crud.Factory;
+import com.redhat.lightblue.metadata.*;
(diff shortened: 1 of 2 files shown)


luan-cestari:
Depends on https://github.com/lightblue-platform/lightblue-core/pull/344

metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
```diff
@@ -137,12 +102,10 @@ public EntityMetadata getEntityMetadata(String entityName,
             }
             return new EntityMetadata(info, schema);
         } catch (Error | IllegalArgumentException e) {
-            // rethrow lightblue error or illegao arg exception
+            // rethrow lightblue error or IllegalArgumentException
             throw e;
         } catch (Exception e) {
-            // throw new Error (preserves current error context)
-            LOGGER.error(e.getMessage(), e);
```
jewzaam: Logging was removed here.

metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
```diff
@@ -167,9 +130,7 @@ public EntityInfo getEntityInfo(String entityName) {
             // rethrow lightblue error
             throw e;
         } catch (Exception e) {
-            // throw new Error (preserves current error context)
-            LOGGER.error(e.getMessage(), e);
```
jewzaam: Logging was removed here.

luan-cestari: Because analyzeException method dos already log that (see first line of analyzeException method https://github.com/luan-cestari/lightblue-mongo/blob/errorCodeTimeoutrest%2312/metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java#L663 )

metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java
luan-cestari: Because analyzeException method dos already log that (see first line of analyzeException method https://github.com/luan-cestari/lightblue-mongo/blob/errorCodeTimeoutrest%2312/metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java#L663 )

jewzaam:
Logging in class `MongoMetadata` was dropped in most cases when switching to use `analyzeException`.

Method `analyzeException` is pretty much the same other than the constants between  `MongoMetadata` and `MongoCRUDController`.  Any thoughts on not duplicating this logic?

PR #92, opened 2015-03-26
Fixes #89 - Mongo test controller

### LINKED ISSUE #89: Create a Mongo pre-enabled AbstractCRUDTestController
This is from a conversation with @paterczm that happened over in: https://github.com/lightblue-platform/lightblue-test-utils/pull/6

Essentially the idea is that we want to make a nice fluid experience for mongo backend users as that is the common (only) use case and we want it to be as simple as possible.

dcrissman:
Adds an implementation of AbstractCRUDTestController that has Mongo support built in, as that is the only supported metadata backend at this time.

### DIFF
--- a/test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
+++ b/test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
@@ -1,3 +1,21 @@
+/*
+ Copyright 2015 Red Hat, Inc. and/or its affiliates.
+
+ This file is part of lightblue.
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package com.redhat.lightblue.mongo.test;
 
 import static com.redhat.lightblue.util.JsonUtils.json;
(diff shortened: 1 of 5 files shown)


dcrissman:
Fixes https://github.com/lightblue-platform/lightblue-mongo/pull/92

dcrissman:
Fixes #89

dcrissman:
This change depends on https://github.com/lightblue-platform/lightblue-core/pull/356

dcrissman:
Ready for review

test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public AbstractMongoCRUDTestController() throws Exception {
+        super(true);
+    }
+
+    @Override
+    protected JsonNode getDatasourcesJson() {
+        try {
+            if (getDatasourcesResourceName() == null) {
+                return json(loadResource("/mongo-datasources.json", true));
```
jewzaam: Should the resource be `./mongo-datasources.json` instead?

test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
```diff
@@ -0,0 +1,100 @@
...
+    @BeforeClass
+    public static void prepareMongoDatasources() {
+        if (System.getProperty("mongo.host") == null) {
+            System.setProperty("mongo.host", "localhost");
+        }
+        if (System.getProperty("mongo.port") == null) {
+            System.setProperty("mongo.port", String.valueOf(mongoServer.getPort()));
+        }
+        if (System.getProperty("mongo.database") == null) {
+            System.setProperty("mongo.database", "lightblue");
```
jewzaam: Why pick a default database name of "lightblue"?  I think we've used "mongodata" or "data" elsewhere, so curious.

dcrissman: It is the value I have used elsewhere, but really I just threw the value in. I am happy to use something else if you'd like.

test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
dcrissman: Honestly, I am not sure. Both ways seem to work. I followed @paterczm (name drop alert) example  over in https://github.com/lightblue-platform/lightblue-core/blob/master/test/src/main/java/com/redhat/lightblue/test/AbstractCRUDTestController.java#L108

I am happy to go either direction.

jewzaam: Guess it doesn't matter if it works.  I've always done the `./` prefix.

PR #95, opened 2015-04-01
Make the backend configurable.

dcrissman:
Takes advantage of changes made in https://github.com/lightblue-platform/lightblue-core/pull/361

Also a dependency for https://github.com/lightblue-platform/lightblue-migrator/issues/142

### DIFF
--- a/test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
+++ b/test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
@@ -64,6 +64,11 @@ public abstract class AbstractMongoCRUDTestController extends AbstractCRUDTestCo
     }
 
     @Override
+    protected JsonNode getLightblueMetadataJson() throws Exception {
+        return json(loadResource("/mongo-lightblue-metadata.json", true));
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


test/src/main/java/com/redhat/lightblue/mongo/test/AbstractMongoCRUDTestController.java
```diff
@@ -45,15 +45,18 @@
 
     @BeforeClass
     public static void prepareMongoDatasources() {
+        if (System.getProperty("mongo.backend") == null) {
+            System.setProperty("mongo.backend", "mongo");
```
jewzaam: Making this configurable makes no sense.  For mongo, it will always be the string "mongo".

test/src/main/resources/mongo-lightblue-metadata.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type" : "com.redhat.lightblue.mongo.config.MongoMetadataConfiguration",
+    "dataSource" : "${mongo.backend}",
```
jewzaam: backend != datasource

dcrissman:
@jewzaam I believe I was simply using the wrong label. Corrections have been made.

dcrissman:
Ready for review

PR #109, opened 2015-04-29
fixes #107 - remove log from translator

### LINKED ISSUE #107: Remove logging in/out in Translator
https://github.com/lightblue-platform/lightblue-mongo/blob/42f88d2d45a83d096cc308dabc5b386053d90698/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java#L245

Data shouldn't be logged anywhere.

@bserdar is there a reason to keep this in?  Per git, you added this change about a week ago.

dcrissman:
Fixes #107

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
@@ -242,7 +242,6 @@ public class Translator {
         }
         JsonDoc doc = toJson(object, md);
         LOGGER.debug("toJson() return");
-        LOGGER.debug("toJson: in: {}, out: {}",object,doc);
         return doc;
     }
 


bserdar:
These logs are valuable debugging tools. Instead of removing them, either:
  1) configure the logger so DEBUG level logs are written somewhere else, or
  2) we can use a separate logger for logs containing data

dcrissman:
@jewzaam do you want to revert the other related changes then and go with more of what @bserdar is suggesting?

dcrissman:
@jewzaam is this still a change you want? If not, I am going to close it.

jewzaam:
Let's close it and address some standards for logging.  There's an internal doc I'll link on IRC that we could groom and then publish on our developer guide.

PR #116, opened 2015-05-12
fix broken build

dcrissman:
Not sure how/why the master build broke, but this fixes it (and a little bit of cleanup).

### DIFF
--- a/hystrix/pom.xml
+++ b/hystrix/pom.xml
@@ -24,16 +24,17 @@
         <artifactId>lightblue-mongo-pom</artifactId>
         <version>1.5.0-SNAPSHOT</version>
     </parent>
-    <groupId>com.redhat.lightblue.mongo</groupId>
     <artifactId>lightblue-mongo-hystrix</artifactId>
     <packaging>jar</packaging>
-    <version>1.5.0-SNAPSHOT</version>
     <name>lightblue-mongo: ${project.groupId}|${project.artifactId}</name>
     <dependencies>
         <dependency>
+            <groupId>com.netflix.hystrix</groupId>
+            <artifactId>hystrix-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.redhat.lightblue.mongo</groupId>
             <artifactId>lightblue-mongo-test</artifactId>
-            <version>1.5.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -147,12 +147,23 @@
...
+            </dependency>
+            <dependency>
                 <groupId>com.redhat.lightblue</groupId>
                 <artifactId>lightblue-core-test</artifactId>
                 <version>1.5.0-SNAPSHOT</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
+                <groupId>com.redhat.lightblue.mongo</groupId>
+                <artifactId>lightblue-mongo-test</artifactId>
```
jewzaam: An undocumented pattern we've been following is to only put dependencies outside of the repo in dependencyManagement.  In my mind it's defining an artifact that doesn't exist yet doing this, but it technically works.

hystrix/pom.xml
```diff
@@ -24,16 +24,17 @@
         <artifactId>lightblue-mongo-pom</artifactId>
         <version>1.5.0-SNAPSHOT</version>
     </parent>
-    <groupId>com.redhat.lightblue.mongo</groupId>
     <artifactId>lightblue-mongo-hystrix</artifactId>
     <packaging>jar</packaging>
-    <version>1.5.0-SNAPSHOT</version>
```
jewzaam: Why is groupId and version removed?

pom.xml
dcrissman: I was unaware of this pattern, but that said, the dependency isn't realized until a child module defines it in its dependencies.

bserdar: This ties the version of the referenced project to the version of the root pom. Those two things should not be tied together. Allows for creating a dependency in a sub-project to a version of another entity whose version number is determined by the root pom version number. Please remove this.

hystrix/pom.xml
dcrissman: These are provided via the parent pom and maven considers them to be redundant.

pom.xml
dcrissman: Reverted in #117

hystrix/pom.xml
jewzaam: This assumes versions are consistent within a repo.  If that assumption is made then #117 shouldn't have been necessary.  I would like this to be added back in for two reasons: 1) it's explicit, 2) we know it works with the release scripts.  If we decide that all versions will be identical in all modules of a repo then we could remove this assuming the release script works, but I don't think we've explicitly had that discussion yet.

dcrissman: Reverted in #118

PR #130, opened 2015-07-09
Put all errors into the context instead of throwing them, so they won…

bserdar:
…'t be lost

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/MongoCRUDController.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/MongoCRUDController.java
@@ -180,11 +180,10 @@ public class MongoCRUDController implements CRUDController, MetadataListener {
                 ctx.getHookManager().queueHooks(ctx);
             }
         } catch (Error e) {
-            // rethrow lightblue error
-            throw e;
+            ctx.addError(e);
         } catch (Exception e) {
             LOGGER.error("Error during insert: {}", e);
-            throw analyzeException(e, CrudConstants.ERR_CRUD);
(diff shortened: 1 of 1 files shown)


crud/src/main/java/com/redhat/lightblue/crud/mongo/MongoCRUDController.java
```diff
@@ -349,11 +347,10 @@ public CRUDFindResponse find(CRUDOperationContext ctx,
                 ctx.addError(Error.get(MongoCrudConstants.ERR_NO_ACCESS, "find:" + ctx.getEntityName()));
             }
         } catch (Error e) {
-            // rethrow lightblue error
-            throw e;
+            ctx.addError(e);
```
paterczm: Why isn't the error logged (LOGGER.error) here? It's very helpful to have all errors logged lightblue side, clients may not log them.

bserdar: We have logging in some of the Error.get methods. Lets add logging to all
Error.get methods, so these would be logged everywhere.

paterczm: Are Error.get always called when an error happens? If so and they are all supposed to log, then statements
   LOGGER.error(e.getMessage(), e);
in catch are redundant.

bserdar: Yes, all Error's are propagated back to client. Anything that is internal
to the system is throw using regular exceptions, and if they propagate,
they are eventually put into an Error.

I'll remove the redundant logs, and add logging to Error.get

PR #135, opened 2015-07-30
Synchronization extension implementation

bserdar:


crud/src/test/java/com/redhat/lightblue/crud/mongo/AbstractMongoCrudTest.java
```diff
@@ -79,8 +80,8 @@ public void setup() throws Exception {
 
     @After
     public void teardown() throws Exception {
-        mongo.getConnection().dropDatabase(DATABADSE_NAME);
-        coll = null;
+        if(coll!=null)
+            coll.remove(new BasicDBObject());
```
dcrissman: Not sure I understand this change. Why is this better?

bserdar: Dropping the database is done asynchronously. So after dropDatabase
returns, the next test starts running, and then the actual drop takes
place, screwing it up.

Removing eveything is synchronous.

dcrissman: Do we still want to set coll = null on the way out?

bserdar: No need. Collection still exists.

crud/src/main/java/com/redhat/lightblue/crud/mongo/MongoCRUDController.java
```diff
@@ -366,6 +380,14 @@ public void updatePredefinedFields(CRUDOperationContext ctx, JsonDoc doc) {
     }
 
     @Override
+    public Extension getExtensionInstance(Class<? extends Extension> extensionClass) {
+        if(extensionClass.equals(LockingSupport.class))
+            return new MongoLockingSupport(this);
+        else
+            return null;
```
dcrissman: Should it throw an exception, or maybe log a warning?

bserdar: No, API spec says if it doesn't support it, it returns null.

Not all CRUD implementations will support all extensions.

crud/src/test/java/com/redhat/lightblue/crud/mongo/AbstractMongoCrudTest.java
dcrissman: Ok, I am good with that then.

dcrissman:
Moving the configuration files will likely cause problems for test cases in downstream projects.

dcrissman:
I created PRs for the downstream projects that I found that need new config files. I will go ahead and merge this PR and it is easy enough to see if anything was missed.

PR #145, opened 2015-08-27
Snapshot version of metadata overwrites old snapshot

### LINKED ISSUE #140: We should allow SNAPSHOT metadata to overwrite existing versions

bserdar:
Fixes #140


dcrissman:
Does this mean that only 1 snapshot version can exist at a time?

bserdar:
No. 1.2.SNAPSHOT and 1.3.SNAPSHOT can exist at the same time. If you create
a new 1.3.SNAPSHOT, it overwrites the existing 1.3.SNAPSHOT, instead of
failing.

PR #148, opened 2015-09-08
Add metadata caching

bserdar:


metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MetadataCache.java
```diff
@@ -0,0 +1,180 @@
...
+        EntityVersion v=new EntityVersion(entityName,version);
+        WeakReference<EntityMetadata> ref=cache.get(v);
+        if(ref!=null)
+            md=ref.get();
+        else
+            md=null;
+
+        long now=System.currentTimeMillis();
+        if(lastCacheRefreshTime+cacheTTLMsecs<now)
+            fullRefresh(collection,now);
```
jewzaam: Why isn't full refresh being done before getting something from the cache?  At this point we'll have stale metadata, at least it's at the edge of the TTL for the cache.

bserdar: Leftover artifact from an earlier implementation. Fixed now.

PR #153, opened 2015-09-15
Dont project refs

bserdar:


jewzaam:
@bserdar If I remove the Translator change the unit tests all pass.  I would expect it to fail if there was something to fix.  Am I missing something?

bserdar:
You are right. Looking...

crud/src/main/java/com/redhat/lightblue/crud/mongo/Translator.java
```diff
@@ -380,21 +381,31 @@ public DBObject translate(EntityMetadata md, UpdateExpression expr)
...
-            }
+        if(cursor.next()) {
+            boolean done=false;
+            do {
+                Path field=cursor.getCurrentPath();
+                FieldTreeNode node=cursor.getCurrentNode();
+                LOGGER.debug("Checking if {} is included ({})",field,node);
+                if(node instanceof ResolvedReferenceField||
+                   node instanceof ReferenceField) {
+                    done=!cursor.nextSibling();
```
jewzaam: Couldn't this stop too early?  What if the ref field is in the middle of the cursor?  What if there are multiple references in the query?

bserdar: It shouldn't stop early. cursor.next moves depth first. When we see a
resolvedReference, it calls nextSibling, i.e. should skip the whole subtree
rooted at that reference, and continue depth first.

PR #155, opened 2015-09-16
use the same datasource as the system properties

dcrissman:
Fixes an issue where the mongo datasource could be different between the lightblue-metadata.json and the test metadata.


paterczm:
Merge when build passes.

PR #160, opened 2015-09-17
Upsert with _id lookup inserts because _id is always null in lookup

bserdar:


jewzaam:
Not test for this?

bserdar:
Existing test tests it. Once I add identity constraint to the test data, it
started failing.

PR #161, opened 2015-09-17
_id field is not added with identity constraint any more

bserdar:


crud/src/test/java/com/redhat/lightblue/crud/mongo/MongoCRUDControllerTest.java
```diff
@@ -1068,8 +1068,6 @@ public void ensureIdFieldTest() throws Exception {
         SimpleField id=(SimpleField)e.getFields().getField("_id");
         Assert.assertNotNull(id);
         Assert.assertEquals(StringType.TYPE,id.getType());
-        Assert.assertEquals(1,id.getConstraints().size());
```
jewzaam: Could we verify there are no constraints instead of removing this row?

bserdar: Done.

PR #172, opened 2015-10-15
Fixes #48 - Configurable WriteConcern

### LINKED ISSUE #48: Write concern needs to be configurable
Right now we use SAFE which doesn't verify data is flushed to disk.  Two problems:
1. it's hard coded, we have no option for changing it in configuration
2. SAFE is probably not a great idea when we cannot afford to loose data

I propose we make it configurable in `datasources.json` per datasource and review what is the best for our deployment.  

http://api.mongodb.org/java/2.6/com/mongodb/WriteConcern.html

I say we configure for FSYNC_SAFE until we have a larger replica set.

BVulaj:


crud/src/main/java/com/redhat/lightblue/crud/mongo/MongoSequenceGenerator.java
```diff
@@ -88,7 +87,7 @@ public long getNextSequenceValue(String name,long init,long inc) {
                 append(INC,inc).
                 append(VALUE,init);
             try {
-                coll.insert(u,WriteConcern.SAFE);
+                coll.insert(u);
```
bserdar: Put SAFE back here. This needs to be safe.

config/src/main/java/com/redhat/lightblue/mongo/config/MongoConfiguration.java
```diff
@@ -238,6 +240,7 @@ public MongoClientOptions getMongoClientOptions() {
             // construct a URI to connect
             builder.socketFactory(getSocketFactory());
         }
+        builder.writeConcern(writeConcern);
```
bserdar: Don't set the write concern for all calls. Synchronization sensitive calls use the same data sources. Instead, pass the write concern to the hystrix command, and use write concern for each call.

PR #173, opened 2015-10-15
Increse wait time for mongo tear down to ensure test stability

paterczm:
Ugly, but will do for now.


bserdar:
This sucks. There should be a better way of doing this.

paterczm:
> This sucks. There should be a better way of doing this.

I logged it as an issue to look at later: https://github.com/lightblue-platform/lightblue-mongo/issues/174

PR #175, opened 2015-10-16
changes to allow for no upper bound

snehagunta:


bserdar:
Also pls make sure this is treated correctly in json parsing/conversion
code in core, in FindRequest class.

PR #177, opened 2015-10-16
range no upper bound

snehagunta:
changed find function in mongo to accommodate unbounded range. 
modified cursor and added tests for [x, null], [0,0]

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/mongo/BasicDocFinder.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/mongo/BasicDocFinder.java
@@ -63,14 +63,23 @@ public class BasicDocFinder implements DocFinder {
             cursor = cursor.sort(mongoSort);
             LOGGER.debug("Result set sorted");
         }
-        long ret = cursor.size();
+        int size = cursor.size();
+        long ret = size;
         LOGGER.debug("Applying limits: {} - {}", from, to);
         if (from != null) {
             cursor.skip(from.intValue());
+            if(to!=null && from>to)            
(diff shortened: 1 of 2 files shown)


crud/src/main/java/com/redhat/lightblue/crud/mongo/BasicDocFinder.java
```diff
@@ -63,14 +63,23 @@ public long find(CRUDOperationContext ctx,
             cursor = cursor.sort(mongoSort);
             LOGGER.debug("Result set sorted");
         }
-        long ret = cursor.size();
+        int size = cursor.size();
+        long ret = size;
         LOGGER.debug("Applying limits: {} - {}", from, to);
         if (from != null) {
             cursor.skip(from.intValue());
+            if(to!=null && from>to)
```
bserdar: Add some comments here to clarify that this means no results will be returned.

crud/src/main/java/com/redhat/lightblue/crud/mongo/BasicDocFinder.java
```diff
@@ -63,14 +63,23 @@ public long find(CRUDOperationContext ctx,
...
             cursor.skip(from.intValue());
+            if(to!=null && from>to)            
+            	cursor.skip(size);	        
         }
+        
         if (to != null) {
+        	if(to >= 0)
             cursor.limit(to.intValue() - (from == null ? 0 : from.intValue()) + 1);
+        	else if(to < 0)
+        		cursor.skip(size);
```
bserdar: Same here, some comments to explain what's going on

PR #185, opened 2015-12-01
Add limit to finder (hardcoded)

bserdar:


jewzaam:
I'm concerned that this logs the raw query to the server.log.  Can we put it somewhere else?  If the query includes sensitive info it would get fed to log aggregator and be available to anybody with access.  If this could be put in a file that isn't indexed it would be better.

bserdar:
It only prints resultsetsize to server log as warning now, and prints prints the query as debug to RESULTSET_LOGGER.

PR #187, opened 2015-12-01
Log only slow results

bserdar:


crud/src/main/java/com/redhat/lightblue/crud/mongo/BasicDocFinder.java
```diff
@@ -98,9 +98,9 @@ else if(to < 0)
         LOGGER.debug("Retrieved {} results", mongoResults.size());
         List<JsonDoc> jsonDocs = translator.toJson(mongoResults);
 
-        if(RESULTSET_LOGGER.isInfoEnabled()) {
-            RESULTSET_LOGGER.info("Execution time:{}, Retrieval time:{}, Resultset size: {}, Data size:{}, Query:{}",
-                                  executionTime,retrievalTime,mongoResults.size(),Translator.size(jsonDocs),mongoQuery);
+        if(RESULTSET_LOGGER.isDebugEnabled()&&(executionTime>100 || retrievalTime>100)) {
+            RESULTSET_LOGGER.debug("Execution time:{}, Retrieval time:{}, Resultset size: {}, Data size:{}, Query:{}",
```
paterczm: Can you make the log format splunk friendly, i.e. execution_time={}, retrieval_time={}...? That way, if we decide to log this to server.log (or other splunk indexed file), we could query with conditions, sort, etc.

Issue #193, opened 2015-12-21
Merge cannot merge nested objects inside arrays


bserdar:
Merge is descending the tree to find the parent of the field to be merged, and trying to find an identity there. Logic fails if there's an array of objects, and a nested object inside an array element, and the field to be merged is there.

bserdar:
Related to core #546

bserdar:
Fixed by #204

PR #204, opened 2016-01-20
199 193

### LINKED ISSUE #199: Access checking during save fails when array elements move
It is comparing elements in array1[i] with arrat2[i], so when an array element is inserted or removed, all elements shift, failing the test

bserdar:
Fix #199 and #193


mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    @Override
+    protected DBObject asObject(Object value) {
+        return (DBObject)value;
+    }
+
+    @Override
+    protected boolean equals(Object value1,Object value2) {
+        return (value1==null&&value2==null) ||
```
BVulaj: First check seems redundant.

bserdar: why?

BVulaj: Woops, misread it.  A-OK.

PR #208, opened 2016-01-22
Various fixes for dereference null pointer and resource leaks.

jewzaam:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -266,9 +266,9 @@ public static Object getDBObject(DBObject start, Path p) {
             String segment = p.head(seg);
             if (segment.equals(Path.ANY)) {
                 throw Error.get(MongoCrudConstants.ERR_TRANSLATION_ERROR, p.toString());
-            } else if (Util.isNumber(segment)) {
+            } else if (trc != null && Util.isNumber(segment)) {
```
bserdar: trc can't be null here, check is unnecessary

jewzaam: It depends on the start param.  Would it not be good to check for it here?  If we don't do this check, some day if null is passed it will blow up with NPE.

PR #212, opened 2016-02-02
Fixes #206 - Case Insensitive Index Support

### LINKED ISSUE #206: Add case insensitive indexes

BVulaj:
Please do not merge yet as this requires some script additions for mongo.


BVulaj:
See: http://dev.forum.lightblue.io/Case-insensitive-search-support-td409.html

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -658,8 +668,13 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
...
                 LOGGER.info("Creating index {} with {}",index.getName(),index.getFields());
                 DBObject newIndex = new BasicDBObject();
-                for (SortKey p : index.getFields()) {
-                    newIndex.put(p.getField().toString(), p.isDesc() ? -1 : 1);
+                for (IndexSortKey p : index.getFields()) {
+                    String field = p.getField().toString();
+                    if (p.isCaseInsensitive()) {
+                        // if this is a case insensitive key, we need to change the field to how mongo actually stores the index
+                        field = Translator.getHiddenForField(p.getField()).toString();
+                    }
```
bserdar: p.getField().toString() is not correct. Instead, change getHiddenForfield accept a Path, not String, and use Translator.translatePath() to translate path.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -692,22 +707,29 @@ private DBObject findIndexWithSignature(List<DBObject> existingIndexes,Index ind
...
             return true;
         return false;
     }
+    
+    private boolean compareSortKeys(IndexSortKey sortKey, String fieldName, Object dir) {
+        String field = sortKey.getField().toString();
+        if (sortKey.isCaseInsensitive()) {
+            // if this is a case insensitive key, we need to change the field to how mongo actually stores the index
+            field = Translator.getHiddenForField(sortKey.getField()).toString();
+        }
```
bserdar: Don't use field.toString()

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -431,6 +433,29 @@ public DBObject translateProjection(EntityMetadata md,
...
+     * @param dbo
+     * @return
+     */
+    public static DBObject populateHiddenFields(EntityMetadata md, DBObject dbo) {
+        md.getEntityInfo().getIndexes().getCaseInsensitiveIndexes().stream()
+            .map(IndexSortKey::getField)
+            .forEach(p -> {
+                Path hiddenPath = getHiddenForField(p);
+                Object object = dbo.get(p.toString());
+                if(object instanceof String) {
```
bserdar: This won't work if p has multiple segments. You need to descend down the dbo tree. There should be a function somewhere in translator that does that

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -659,7 +691,31 @@ private DBObject translateRegexMatchExpression(RegexMatchExpression expr) {
...
+     * Get a reference to the path's hidden sub-field.
+     *
+     * This does not guarantee the sub-path exists.
+     *
+     * @param path
+     * @return
+     */
+    public static Path getHiddenForField(Path path) {
+        return path.prefix(-1).mutableCopy().push(HIDDEN_SUB_PATH).push(path.getLast());
+    }
```
bserdar: This isn't well defined if path itself is an instance of MutablePath, because then mutableCopy() returns this. Store path.getLast() in a local var first, then modify it.

bserdar:
You need to add tests for:
  - Modifying an existing index to case insensitive should drop existing index, and create the new one
  - Creating new case insensitive index
  - Modifying existing case insensitive index to case sensitive
  - Drop case insensitive index

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
BVulaj: @bserdar getHiddenForField actually already accepts a Path, not a String.  What's going on here is it's returning a Path, and then that is being converted to a string.  Is that still incorrect?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -200,6 +208,27 @@ public static String translateJsPath(Path p) {
     }
 
     /**
+     * Return a full field representation of the given index based on the provided entity metadata.  This will include asterisks to represent array indexes.
+     *
+     * @param md
+     * @param index
+     * @return
+     */
+    public static FieldTreeNode getFieldByIndex(EntityMetadata md, Path index) {
```
bserdar: This is wrong. It should look like this:
MutablePath p=new MutablePath()
for(i=0 to numsegments) {
  String seg=index.head(i);
  p.push(seg);
  if(md.resolve(index.prefix(i+1)) is an array)
    p.push(*)
}
return md.resolve(p);

mongo/src/main/resources/js/populate-hidden-fields.js
```diff
@@ -0,0 +1,44 @@
...
+        cursor.forEach(function(doc) {
+            // fieldMap is <index, hiddenField>
+            for (var k in fieldMap) {
+                // check if there's an array in the index
+                var arrayIndex = k.lastIndexOf("*");
+                if (arrayIndex > -1) {
+                    doArrayMap(doc, k, fieldMap[k], arrayIndex);
+                } else {
+                    // if no array, just do a direct update
+                    doc[fieldMap[k]] = doc[k].toUpperCase();
```
bserdar: Does this add the necessary levels if k has multiple levels?

BVulaj: What do you mean by that?

This should be able to handle updating multi-nested arrays.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
BVulaj: Updated.

mongo/src/main/resources/js/populate-hidden-fields.js
bserdar: If fieldMap[k] is x.y.z, and the doc doesn't have x.y, what happens?

BVulaj: Ah, you just mean if the document doesn't have the corresponding field to update?  This will probably fail then.  I'll add in a check.

BVulaj:
Tests have been updated and added as per @bserdar's comments.

dcrissman:
lightblue-mongo is broken until this is merged in

dcrissman:
I re-ran travis, but it seems tests are still failing.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -695,24 +772,169 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
...
+                    ObjectNode arrNode = JsonNodeFactory.instance.objectNode();
+                    JsonDoc.modify(arrNode, new Path(fieldMap.get(index)), JsonNodeFactory.instance.textNode(dbObject.toString().toUpperCase()), true);
+                    ObjectMapper mapper = new ObjectMapper();
+                    JsonNode merged = merge(mapper.readTree(doc.toString()), mapper.readTree(arrNode.toString()));
+                    DBObject obj = (DBObject) JSON.parse(merged.toString());
+                    ((BasicDBObject) doc).clear();
+                    ((BasicDBObject) doc).putAll(obj);
+                }
+            }
+            coll.save(doc);
```
bserdar: Might be nice if you can skip save if doc is not changed.

BVulaj: Agreed.  I'll look into doing that.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -695,24 +773,187 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
...
+     *
+     * @param md
+     * @throws IOException
+     */
+    public void reindex(EntityMetadata md) throws IOException {
+        Map<String, String> fieldMap = Translator.getCaseInsensitiveIndexes(md.getEntityInfo().getIndexes().getIndexes())
+                .collect(Collectors.toMap(
+                        i -> i.getField().toString(),
+                        i -> Translator.getHiddenForField(i.getField()).toString()));
+        populateHiddenFields(md.getEntityInfo(), fieldMap);
```
bserdar: Don't do this if there are no case insensitive indexes

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -695,24 +773,187 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
...
+                    ObjectNode arrNode = JsonNodeFactory.instance.objectNode();
+                    JsonDoc.modify(arrNode, new Path(fieldMap.get(index)), JsonNodeFactory.instance.textNode(dbObject.toString().toUpperCase()), true);
+                    ObjectMapper mapper = new ObjectMapper();
+                    JsonNode merged = merge(mapper.readTree(doc.toString()), mapper.readTree(arrNode.toString()));
+                    DBObject obj = (DBObject) JSON.parse(merged.toString());
+                    ((BasicDBObject) doc).clear();
+                    ((BasicDBObject) doc).putAll(obj);
+                }
+            }
+            if (!doc.equals(original)) {
```
bserdar: Are you sure this works?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -695,24 +773,187 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
...
+                coll.save(doc);
+            }
+        }
+        cursor.close();
+        LOGGER.debug("Finished population of hidden fields.");
+    }
+
+    private void doMultiArrayMap(DBObject doc, String index, String hidden, int arrIndex) throws IOException {
```
bserdar: Some comments here would be nice. wth does it "do"?

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -90,6 +102,250 @@ public DB get(MongoDataStore store) {
...
+
+        List<Index> indexes = new ArrayList<>();
+        indexes.add(index1);
+        indexes.add(index2);
+        indexes.add(index3);
+        indexes.add(index4);
+        e.getEntityInfo().getIndexes().setIndexes(indexes);
+        return e;
+    }
+
```
bserdar: Add a test that verifies an already existing index is not recreated upon metadata update

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
BVulaj: @bserdar Based from the doc http://api.mongodb.org/java/current/org/bson/BasicBSONObject.html#equals-java.lang.Object-

I would expect it to.  I can try and add a test to verify.

bserdar: Doc says so, no need to verify

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
BVulaj: Cleaning this up a bit.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -683,11 +746,26 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei) {
                 LOGGER.info("Dropping {}",index.get("name"));
                 entityCollection.dropIndex(index.get("name").toString());
             }
+            boolean background = true;
```
bserdar: 'background' is misleading. What you're really doing here is determining if a new case insensitive index is to be created. Rename the flag to signify that, and add a comment where i tis used in place of background option.

PR #217, opened 2016-02-23
216

bserdar:


jewzaam:
@bserdar could you merge master?  I've opened https://github.com/lightblue-platform/lightblue-core/issues/606 to capture (and discuss) the desire to have the preference generalized.  We need this capability so will not hold it up, we can update things later.

PR #229, opened 2016-03-23
fix compilation error

dcrissman:
Switched to using the new IndexSortKey subclass of SortKey.


dcrissman:
This appears to be addressed in https://github.com/lightblue-platform/lightblue-mongo/pull/212. I am going to close this PR.

PR #232, opened 2016-03-30
New BSONParser implementation using new metadata parser methods

bserdar:
Depends on https://github.com/lightblue-platform/lightblue-core/pull/624

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
@@ -56,7 +56,7 @@ public class BSONParser extends MetadataParser<Object> {
 
     public static final String DELIMITER_ID = "|";
 
-    public BSONParser(Extensions ex,
+    public BSONParser(Extensions<Object> ex,
                       TypeResolver resolver) {
         super(ex, resolver);
     }
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
@@ -51,140 +51,86 @@ import com.redhat.lightblue.util.Error;
 import com.redhat.lightblue.util.JsonUtils;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class BSONParser extends MetadataParser<BSONObject> {
+public class BSONParser extends MetadataParser<Object> {
     private static final Logger LOGGER = LoggerFactory.getLogger(BSONParser.class);
 
     public static final String DELIMITER_ID = "|";
 
-    public BSONParser(Extensions<BSONObject> ex,
+    public BSONParser(Extensions ex,
                       TypeResolver resolver) {
         super(ex, resolver);
     }
 
     @Override
-    public String getStringProperty(BSONObject object, String name) {
-        Object x = object.get(name);
-        if (x != null) {
-            if (x instanceof String) {
-                return (String) x;
-            } else {
-                throw Error.get(MetadataConstants.ERR_ILL_FORMED_METADATA, name);
-            }
-        } else {
-            return null;
-        }
-    }
-
-    @Override
-    public BSONObject getObjectProperty(BSONObject object, String name) {
-        Object x = object.get(name);
-        if (x != null) {
-            if (x instanceof BSONObject) {
-                return (BSONObject) x;
-            } else {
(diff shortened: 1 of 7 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -51,140 +51,86 @@
...
-                }
-                return ret;
-            } else {
-                throw Error.get(MetadataConstants.ERR_ILL_FORMED_METADATA, name);
-            }
-        } else {
-            return null;
-        }
+    public Object newMap() {
+        return new BasicDBObject();
```
dcrissman: You could specifically return BasicDBObject here and still conform to the Object generic type. Might save unnecessary casting elsewhere?

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -51,140 +51,86 @@
...
+    public void setMapProperty(Object map,String name,Object value) {
+        ((BSONObject)map).put(name,value);
     }
 
-    @Override
-    public void putString(BSONObject object, String name, String value) {
-        object.put(name, value);
+   @Override
+    public  Object newList() {
+       return new ArrayList<Object>();
```
dcrissman: Could return type List<Object> and still conform to the generic type.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
bserdar: How? This overrides a T newMap(), and T is bound to Object.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -51,140 +51,86 @@
...
 
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class BSONParser extends MetadataParser<BSONObject> {
+public class BSONParser extends MetadataParser<Object> {
     private static final Logger LOGGER = LoggerFactory.getLogger(BSONParser.class);
 
     public static final String DELIMITER_ID = "|";
 
-    public BSONParser(Extensions<BSONObject> ex,
+    public BSONParser(Extensions ex,
```
dcrissman: Should include generic type of Object

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
dcrissman: BasicDBObject is an Object, so it still conforms.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
bserdar: Fixed.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
bserdar: But what's the point? It won't save any typecasts,. newMap() is not used
here, it is used in MetadataParser, and there, it uses type T.

dcrissman: I won't block on this, but as a rule of thumb, I try to return Objects at the highest level possible and pass in parameters at the lowest level as possible. In this case, if there were a use case to use BSONParser directly, then casting wouldn't be required.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
dcrissman: Won't block on this either.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
bserdar: Yes, but the metadataparser was incorrectly parameterized. It takes only
one parameter <T>, and assumes the metadata is represented using objects
all deriving from T. This is correct for JSON, where everyrhing is a
JsonNode, but it is not correct for BSON, because maps are DBObject, values
are Java values and lists are java Lists. So, Object is the only common
denominator for BSON.

The parser should have been parameterized using <L, M, V>, different types
for list, map, and value.

PR #237, opened 2016-04-06
Added a more complex insert test

BVulaj:
Also modified translator.getDbObject to return null if a field does not exist, rather than throwing an error

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
@@ -282,7 +282,8 @@ public class Translator {
                 trc = ((DBObject) trc).get(segment);
             }
             if (trc == null&&seg+1<n) {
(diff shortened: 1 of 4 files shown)


BVulaj:
@dcrissman Here is a test that **should** replicate what you were doing in dev, but this test passes without issue.  Take a look and see if there's anything different between what's going on here and what's happening in prod.  FWIW - I just pulled the data from dev

dcrissman:
Below is the exception that I am getting, however when I step through the new unit test BasicDBObject is the DBObject implementation that is used. I am not sure what accounts for the implementation difference.

```java
java.lang.IllegalArgumentException: BasicBSONList can only work with numeric keys, not: [emails]
	at org.bson.types.BasicBSONList._getInt(BasicBSONList.java:168)
	at org.bson.types.BasicBSONList._getInt(BasicBSONList.java:160)
	at org.bson.types.BasicBSONList.get(BasicBSONList.java:105)
	at com.redhat.lightblue.mongo.crud.Translator.getDBObject(Translator.java:282)
	...
```

PR #242, opened 2016-04-11
Fixed some SortKey -> IndexSortKey instances

BVulaj:
Also added a test for arrays nested in arays


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -556,11 +555,11 @@ private void validateIndexFields(EntityInfo ei) {
             List<IndexSortKey> newFields = null;
             boolean copied = false;
             int i = 0;
-            for (SortKey key : fields) {
+            for (IndexSortKey key : fields) {
                 Path p = key.getField();
                 Path newPath = translateIndexPath(p);
                 if (!p.equals(newPath)) {
-                    IndexSortKey newKey = new IndexSortKey(newPath, key.isDesc());
+                    IndexSortKey newKey = new IndexSortKey(newPath, key.isDesc(), key.isCaseInsensitive());
```
BVulaj: @dcrissman @bserdar I think this might have been affecting the index creation.

PR #246, opened 2016-04-13
Removed index validation method which unnecessarily translated indexes

BVulaj:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -479,7 +479,6 @@ public void afterUpdateEntityInfo(Metadata md, EntityInfo ei, boolean newEntity)
 
     @Override
     public void beforeUpdateEntityInfo(Metadata md, EntityInfo ei, boolean newEntity) {
```
bserdar: Remove the method as well

PR #250, opened 2016-04-20
Added a fix so updates will modify the hidden fields as well

BVulaj:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -1258,11 +1260,12 @@ private BasicDBObject objectToBson(JsonNodeCursor cursor, EntityMetadata md) {
...
             JsonNode node = cursor.getCurrentNode();
             LOGGER.debug("field: {}", path);
-            FieldTreeNode fieldMdNode = md.resolve(path);
-            if (fieldMdNode == null) {
-                throw Error.get(ERR_INVALID_FIELD, path.toString());
+            FieldTreeNode fieldMdNode = null;
+            try {
+                fieldMdNode = md.resolve(path);
+            } catch (Exception e) {
+                return null;
```
bserdar: This hides the exception. I think removing the try-catch would be the best,

BVulaj: @bserdar How will we get a null returned then?  Could we just log it instead?  `md.resolve()` won't return a null, it'll just throw the exception.

bserdar: Null won't be returned. The old code checks null and throws an exception. Null check is unnecessary, resolve already throws the exception. so getting rid of the null check in the old code should be sufficient.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -175,8 +175,13 @@ public boolean merge(DBObject oldDoc,DBObject newDoc) {
                     if(hidden) {
                         // Is the removed field in metadata?
                         try {
-                            md.resolve(removedField);
-                            hidden=false;
+                            // don't try to resolve a hidden field
+                            if (!removedField.equals(Translator.HIDDEN_SUB_PATH)) {
```
bserdar: removedField is a path, you should check if HIDDEN_SUB_PATH is one of its segments, equals won't work.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -175,8 +175,13 @@ public boolean merge(DBObject oldDoc,DBObject newDoc) {
                     if(hidden) {
                         // Is the removed field in metadata?
                         try {
-                            md.resolve(removedField);
-                            hidden=false;
+                            // don't try to resolve a hidden field
+                            if (hasHiddenField(removedField)) {
+                                md.resolve(removedField);
```
bserdar: if(!hasHiddenField

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -133,6 +133,33 @@ public void ensureIndexNotRecreated() throws IOException, InterruptedException {
...
+        EntityMetadata emd = addCIIndexes(createMetadata());
+        controller.afterUpdateEntityInfo(null, emd.getEntityInfo(),false);
+        JsonDoc doc = new JsonDoc(loadJsonNode("./testdataCI.json"));
+
+        TestCRUDOperationContext ctx = new TestCRUDOperationContext("testEntity", CRUDOperation.INSERT);
+        ctx.add(emd);
+        ctx.addDocument(doc);
+        controller.insert(ctx, null);
+
+        ctx = new TestCRUDOperationContext("testEntity", CRUDOperation.UPDATE);
```
paterczm: Does it make sense to test other operations as well, e.g. CRUDOperation.SAVE?

PR #252, opened 2016-04-22
Our metadata should be saved with *s, but the actual mongo index should be created without

BVulaj:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -715,15 +715,16 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei, Metadata md) {
                 LOGGER.info("Creating index {} with {}",index.getName(),index.getFields());
                 DBObject newIndex = new BasicDBObject();
                 for (IndexSortKey p : index.getFields()) {
-                    String field = Translator.translatePath(p.getField());
+                    String field = p.getField().toString();
```
bserdar: Declare this as Path field, get rid of toString(), things will simplify going forward

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -133,6 +133,23 @@ public void ensureIndexNotRecreated() throws IOException, InterruptedException {
     }
 
     @Test
+    public void findDocument_CI() throws Exception {
```
bserdar: You need a test for elemMatch. I think one way you can do this is:
 - create a doc with CI index in array element field
 - manually modify array index so array.*.field="someValue" and array.*.@mongoHidden.field="REALVALUE"
 - regex search for "realvalue", it should find it.

PR #255, opened 2016-04-28
Removed dependency on EntityMetadata from reindex()

BVulaj:


BVulaj:
Don't merge yet.  Need to add some extra logging.

PR #259, opened 2016-05-06
Fix 233: Use bulk mongo apis for save and insert

bserdar:


jewzaam:
Why was maxQueryTimeMS removed from BasicDocSaver?

bserdar:
Queries are on _id, or on fields with unique indexes so there is not much
of a chance of getting a collection scan here. If query takes long, it is
most likely due to a temporary i/o wait.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -50,123 +54,311 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BasicDocSaver.class);
 
+    private int batchSize=64;
```
jewzaam: Rational behind 64?  It's hard coded, is it worth exposing and making configurable via puppet?  If yes we could create an issue to fix do that later.

bserdar: No rational, just some arbitrary number. We might make it configurable via
the datasource maybe?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -50,123 +54,311 @@
...
+            for(DocInfo doc:batch) {
+                doc.id=getFieldValues(doc.newDoc,idPaths);
+                if(!isNull(doc.id)) {
+                    idQueries.add(doc.getIdQuery());
+                }
+            }
+            if(!idQueries.isEmpty()) {
+                BasicDBObject retrievalq=new BasicDBObject("$or",idQueries);
+                LOGGER.debug("Existing document retrieval query={}",retrievalq);
+                DBCursor cursor=collection.find(retrievalq,null);
```
jewzaam: Should close the cursor.  Recommend try-with-resource.

bserdar: Closed.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
jewzaam: added https://github.com/lightblue-platform/lightblue-mongo/issues/260 for this

PR #267, opened 2016-05-12
Make sure we close the cursor at the end of the population loop

BVulaj:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -834,10 +834,12 @@ public EntityMetadata getEntityMetadata(String entityName) {
                     coll.save(doc);
                 } catch (Exception e) {
                     LOGGER.error(e.getMessage());
-                    LOGGER.error("Error saving doc:\n{}", doc);
+                    LOGGER.debug("Original doc:\n{}", original);
+                    LOGGER.debug("Error saving doc:\n{}", doc);
                 }
             }
         }
+        cursor.close();
```
bserdar: The cursor iteration block can use a try-catch-log-finally, so we can see errors if read fails.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -825,18 +825,26 @@ public EntityMetadata getEntityMetadata(String entityName) {
...
                 } catch (Exception e) {
+                    // skip the doc if there's a problem, don't outright fail
                     LOGGER.error(e.getMessage());
-                    LOGGER.error("Error saving doc:\n{}", doc);
+                    LOGGER.debug("Original doc:\n{}", original);
+                    LOGGER.debug("Error saving doc:\n{}", doc);
                 }
             }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
bserdar: Log it, then throw

PR #275, opened 2016-05-18
Use mongo bulk updates for update requests

BVulaj:


bserdar:
Please add the logic below to your PR as well:

https://github.com/lightblue-platform/lightblue-mongo/pull/282

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -53,7 +53,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BasicDocSaver.class);
 
-    private int batchSize = 64;
+    private final int batchSize = 64;
```
dcrissman: Should this be configurable?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -48,28 +59,39 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IterateAndUpdate.class);
 
+    private final int batchSize = 64;
```
dcrissman: I see this declared a second time. Maybe it should be defined in a single location?

BVulaj: Three times, actually :(

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
BVulaj: It's definitely a configuration candidate, but @bserdar and I just decided to leave it here for now and move it out in another PR.  Is there a place in LB today to wire up configuration properties?

bserdar: Lets leave it in there for now. There is a ticket to move that to
configuration.

dcrissman: Probably the datasources.json, but I would be fine if we spawned a ticket to address this later as a configuration.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
dcrissman: MongoMetadataConstants maybe? At least then we wouldn't loose them later when we do go to make it configurable.

bserdar: How about we get the batchsize in constructor now, declare a constant in
mongocrudcontroller and pass that? Later we can get that value from the
configuration.

BVulaj: I like it.  Incoming.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +192,63 @@ public void update(CRUDOperationContext ctx,
                 }
                 docIndex++;
             }
+            // if we have any remaining items to update
+            if (numUpdating > 0) {
```
bserdar: This is never zero. numUpdating always increases.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +192,63 @@ public void update(CRUDOperationContext ctx,
...
-        response.setNumMatched(docIndex);
+        response.setNumMatched(numMatched);
+    }
+
+    private void handleBulkWriteError(List<BulkWriteError> errors, List<DocCtx> docs) {
+        docUpdateActual.addAll(docs);
+        for (BulkWriteError e : errors) {
+            DocCtx doc = docs.get(e.getIndex());
+            // remove any docs from the updated list if they had an error
+            docUpdateActual.set(e.getIndex(), null);
```
bserdar: ?

BVulaj: Had initially needed to remove items from the list without affecting indexes, but removed that code before committing.  Will just remove this.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
BVulaj: Resetting it in the above block now.

bserdar:
Can you add one more unit test where an updated doc in the middle of the list fails. You create a unique index on something, and then create a duplicate record with an update to make it fail.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +194,62 @@ public void update(CRUDOperationContext ctx,
...
+        docUpdateActual.addAll(docs);
+        for (BulkWriteError e : errors) {
+            DocCtx doc = docs.get(e.getIndex());
+            if (e.getCode() == 11000 || e.getCode() == 11001) {
+                doc.addError(Error.get("update", MongoCrudConstants.ERR_DUPLICATE, e.getMessage()));
+            } else {
+                doc.addError(Error.get("update", MongoCrudConstants.ERR_SAVE_ERROR, e.getMessage()));
+            }
+        }
+        numFailed += (int) errors.stream().map(e -> e.getIndex()).distinct().count();
```
bserdar: Instead of this, inc numFailed in the loop above.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +194,62 @@ public void update(CRUDOperationContext ctx,
...
+        });
+
         response.setNumUpdated(numUpdated);
         response.setNumFailed(numFailed);
-        response.setNumMatched(docIndex);
+        response.setNumMatched(numMatched);
+    }
+
+    private void handleBulkWriteError(List<BulkWriteError> errors, List<DocCtx> docs) {
+        docUpdateActual.addAll(docs);
```
bserdar: This has all the docs, not just the actual updated ones.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +194,62 @@ public void update(CRUDOperationContext ctx,
...
+        for (BulkWriteError e : errors) {
```
bserdar: You don't really need docupdateActual, ctx has methods that'll give the the list of docs with no errors.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +189,60 @@ public void update(CRUDOperationContext ctx,
...
+                LOGGER.debug("Bulk deleting docs with writeConcern={} from execution", writeConcern);
+                result = bwo.execute(writeConcern);
+            }
+        } catch (BulkWriteException e) {
+            BulkWriteException bwe = e;
+            result = bwe.getWriteResult();
+            LOGGER.warn("Bulk update operation contains errors", bwe);
+            docUpdateErrors.addAll(bwe.getWriteErrors());
+        } catch (Exception e) {
+            throw e;
```
bserdar: Either log it, or don't catch it

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +189,60 @@ public void update(CRUDOperationContext ctx,
                 }
                 docIndex++;
             }
+            // if we have any remaining items to update
+            if (numUpdating > 0) {
+                try {
+                    executeAndLogBulkErrors(bwo);
+                    numFailed += (int) docUpdateErrors.stream().map(e -> e.getIndex()).distinct().count();
```
bserdar: I don't like this. Why not set this it in handleBulkWriteError

BVulaj: This whole block?  Or just the numFailed?  It was there previously but a previous comment asked for me to move it.  I could have misunderstood.

bserdar: numFailed. Can simply do numField++ every time you process an error

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -160,14 +189,58 @@ public void update(CRUDOperationContext ctx,
...
             }
         }
-        response.setNumUpdated(numUpdated);
+        handleBulkWriteError(docUpdateErrors, docUpdateAttempts);
+
+        ctx.getDocumentsWithoutErrors().stream().forEach(doc -> {
+            ctx.getFactory().getInterceptors().callInterceptors(InterceptPoint.POST_CRUD_UPDATE_DOC, ctx, doc);
+        });
+
+        response.setNumUpdated(ctx.getDocumentsWithoutErrors().size());
```
bserdar: getDocumentsWithoutErrors is not numUpdated, it is exactly what it says: docsWithoutErrors, and it is an expensive operation. Instead, set a variable where you process errors.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
BVulaj: The only reason I do this is because it seems possible for > 1 error to be returned per index, which would skew the actual number failed.  Do you think we should just not worry about that case?

BVulaj: If we want to ignore that case, we can simply just use the size of the error and update attempt lists to set numUpdated and numFailed

bserdar: A better way of doing this is to concentrate the logic to a single place:
there is already a function that determines what's an error and what's not.
Count the failed docs there.

PR #276, opened 2016-05-19
Fix #181: Return a different error for the case where upsert=false, b…

### LINKED ISSUE #181: Return a better exception for save with upsert=false, but fails to update
Either the caller is passing in an unique id, or there are no docs with that id. This throws a generic error, it should throw something more meaningful

bserdar:
…ut document is to be inserted


dcrissman:
Needs a rebase if you still want this.

bserdar:
Done/

Issue #280, opened 2016-05-20
Execution writeConcern has no effect


paterczm:
ReadPreference is used at least for some of the operations, but not WriteConcern. The Lightblue client has been recently updated to include read preference, write concern and maxQueryTimeMS execution options.

PR #282, opened 2016-05-20
Adding WriteConcern override support

paterczm:
https://github.com/lightblue-platform/lightblue-mongo/issues/280


bserdar:
Also should be fixed in IterateAndUpdate

bserdar:
Actually, no, lets not fix that in this PR, @BVulaj  is rewriting IterateAndUpdate, we should add this there.

PR #283, opened 2016-05-23
Fix #281: No need to block for index creation for hidden fields

### LINKED ISSUE #281: Index creation for hidden fields should be done in background
https://docs.mongodb.com/manual/core/index-creation/

It locks down everything.

Mongo should be able to deal with modifying docs while the index is being built.

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -785,8 +784,7 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei, Metadata md) {
                 if (index.getName() != null && index.getName().trim().length() > 0) {
                     options.append("name", index.getName().trim());
                 }
-                // if we have hidden fields to generate, we want index creation to be blocking so we can ensure that the indexes are created before we generate the fields
-                options.append("background", !hidden);
+                options.append("background", true);
```
BVulaj: I'm trying to remember - why did we have this blocking in the first place?

bserdar: I don't remember, but we probably thought inserting a doc while building an
index wouldn't work. Why would we have thought that? Maybe things happened?

BVulaj: I'm not sure. Let me think about it before merging.

PR #290, opened 2016-06-10
Fix #289: All queries should have an objectType criteria

### LINKED ISSUE #289: Translator should append objectType to the query

bserdar:


jewzaam:
Will this change lightblue query performance for large collections that do not have `objectType` indexed?

bserdar:
Normally we don't expect this to be indexed, becuse all docs in the
collection will have the same value.  If it is not indexed, it should not
affect performance because mongo will pick index based on other fields.

PR #295, opened 2016-07-29
fixes #294 - remove unneeded and noisy logs and just let there be a debug level log

### LINKED ISSUE #294: TranslationError log noise
Our logs are being blitzed with these two logs over and over again. Is this a legitimate bug or can it be silenced somehow?

```
2016-07-28 20:13:02,178 [https_executor-threads - 265] WARN  [com.redhat.lightblue.mongo.crud.Translator] {"objectType":"error","context":"rest/UpdateCommand/user/update(user:null)/update","errorCode":"mongo-crud:TranslationError","msg":"elm.field1"}
2016-07-28 20:13:49,862 [https_executor-threads - 166] ERROR [com.redhat.lightblue.util.Error] {"objectType":"error","context":"rest/UpdateCommand/user/update(user:null)/update","errorCode":"mongo-crud:TranslationError","msg":"elm.field1"}
```

dcrissman:
Error.get(...) will log the message at an ERROR level, then that message is duplicated at a WARN level. Both are not needed, and I argue neither is needed as explained in #294 this seems to be a legitimate use case.

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
@@ -302,7 +302,7 @@ public class Translator {
                 trc = ((DBObject) trc).get(segment);
             }
             if (trc == null && seg + 1 < n) {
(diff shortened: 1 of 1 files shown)


dcrissman:
@bserdar @BVulaj - Either of you want to weigh in on this PR?

PR #299, opened 2016-08-08
296

bserdar:


dcrissman:
Are you saying that hidden fields will only ever be for case sensitivity? Seems we might want to keep that generic in case we ever need to support some other hidden data.

bserdar:
I am not saying that. What makes you think hidden fields are for case
insensitive indexes only? That should *not* be the case. This change adds a
new hidden field that has got nothing to do with case insensitivity.

dcrissman:
One of the commit messages confused me.

bserdar:
I see, the "Rename..." commit. Yes, there was a method called
populateHiddenFields, and that method is only used for case insensitive
hidden fields, so I renamed that to clarify its purpose.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -63,6 +65,7 @@
     private final Field[] idFields;
     private final Path[] idPaths;
     private final String[] mongoIdFields;
+    private final String txid=ObjectId.get().toString();
```
alechenninger: Some places use txid, some use docver (docversion,etc). I think maybe all of them should just be called docVer/docVersion to be clear they are the same thing.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -288,59 +293,35 @@ private void updateDocs(CRUDOperationContext ctx,
...
                         doc.inputDoc.addError(Error.get("update",
-                                CrudConstants.ERR_NO_FIELD_UPDATE_ACCESS, paths.toString()));
+                                                        CrudConstants.ERR_NO_FIELD_UPDATE_ACCESS, paths.toString()));
                     }
                 }
                 LOGGER.debug("After checks and merge, updating {} docs", updateAttemptList.size());
                 if (!updateAttemptList.isEmpty()) {
-                    BulkWriteOperation bw = collection.initializeUnorderedBulkOperation();
+                    BulkUpdateCtx updateCtx=new BulkUpdateCtx();
+                    updateCtx.reset(collection);
```
alechenninger: Does reset need to be public API? It seems like the contract is only, "call it immediately after you create it," since executing the update resets itself afterward. I think making it private and having the constructor take care of calling it would make it easier to use and less surprising since it wouldn't be constructed starting in an invalid state.

alechenninger: I guess if you did that, it might also make sense to remove collection from the execute param. Is there a reason this is a param? I would think to correctly find any errors, it would have to be the same collection that was passed to reset, anyway? So maybe passing in constructor helps avoid a class of bugs if I understand correctly.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
bserdar: Renamed txid to docver.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
bserdar: Done.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -288,59 +293,34 @@ private void updateDocs(CRUDOperationContext ctx,
...
-                    } catch (BulkWriteException bwe) {
-                        LOGGER.error("Bulk write exception", bwe);
-                        handleBulkWriteError(bwe.getWriteErrors(), "update", updateAttemptList);
-                    } catch (RuntimeException e) {
-                    } finally {
-                        for (DocInfo doc : updateAttemptList) {
-                            if (!doc.inputDoc.hasErrors()) {
-                                ctx.getFactory().getInterceptors().callInterceptors(InterceptPoint.POST_CRUD_UPDATE_DOC, ctx, doc.inputDoc);
-                            }
+                    updateCtx.execute(writeConcern);
```
alechenninger: For save requests, it's interesting that we're detecting concurrent mods too. Is it because it's possible identifying values could have changed between when we retrieved the document to save over and when we actually do the save? Are identity fields allowed to be mutable?

I'm trying to decide if clients issuing a save request would want this behavior or not (and if they do, maybe they really should be doing an update). I guess I expect a save to potentially blow away other concurrent updates since I'm passing in the whole entity.

bserdar: Identifying values won't change, in this context, _id is the only
identifying value, and it is immutable.

I had the same thought while doing this: with a save, concurrent updates
are indistinguishable from serialized updates, one of them wins, and the
result is always consistent. I chose this implementation to notify the side
that lost the race, to force it to reload the latest copy. Maybe that
doesn't really matter?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -106,74 +105,71 @@ public void update(CRUDOperationContext ctx,
         int numMatched = 0;
         int numFailed = 0;
         BsonMerge merge = new BsonMerge(md);
+        BulkUpdateCtx bulkCtx=new BulkUpdateCtx(collection);
         try {
             ctx.getFactory().getInterceptors().callInterceptors(InterceptPoint.PRE_CRUD_UPDATE_RESULTSET, ctx);
             cursor = collection.find(query, null);
```
alechenninger: What would it take to be able to catch concurrent mods and retry from this point, constraining the retry-find to only those document ids which had concurrent mods? To me this would be a killer feature (essentially it would make very expressive updates appear atomic), but I realize maybe hard to implement. Is this something worth considering, if only long term? Or would it be a bad idea?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BulkUpdateCtx.java
```diff
@@ -0,0 +1,185 @@
...
+            if(doc.docCtx.hasErrors())
+                numErrors++;
+        reset();
+    }
+
+    public interface GetCB {
+        DocCtx getCtx(int index);
+    }
+    
+    public static <T> void handleBulkWriteError(List<BulkWriteError> errors,String op,GetCB cb) {
```
alechenninger: nit: unnecessary generic type declaration

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Utils.java
```diff
@@ -0,0 +1,71 @@
...
+     * docversion field of all docs are set to this value for the
+     * current operation
+     * @param idToTxIdMap A mapping of doc _id to the old transaction
+     * id of the document
+     *
+     * @return A set of _ids whose `document version is not equal to currentTxId
+     */
+    public static Set<Object> checkFailedUpdates(DBCollection collection,
+                                                 String currentTxId,
+                                                 Map<Object,String> idToTxIdMap) {
```
alechenninger: nit: a few txid renames missed

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BulkUpdateCtx.java
bserdar: Cleaned.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
bserdar: How many times will we try? Until it is successful?

I think this can be better handled with the solution of the transaction
demarcation problem.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Utils.java
```diff
@@ -0,0 +1,71 @@
...
+        // We would like to find all _id:currentTxId documents. Those
+        // that are missing, i.e. not in the resultset, are the docs
+        // that are modified during or after our update operation
+        DBCursor cursor=null;
+        try {
+            Set<Object> keys=idToTxIdMap.keySet();
+            cursor=collection.find(new BasicDBObject(Translator.DOC_VERSION_FULLPATH_STR,currentTxId).
```
alechenninger: I'm not sure how to provide this feature otherwise, but I think this has a race condition, still, where some updates could be counted as failed even though they actually succeeded. This would happen if, before this find request, another update happened which changed the document version of one of the documents involved in this update. So, I think even though both updates were successful, the earlier one would detect a doc version change since the start of the update transaction, and fail that request.

It's a shame BulkWriteResult doesn't contain more useful info! Is there any way to get the ids of matched documents back from mongo that is still as efficient as BWO?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
alechenninger: There would have to be some reasonable (configurable?) limit to retry attempts. I would have to imagine that would be a rare scenario like some kind of DDOS attack that would cause us to hit a limit, so it probably wouldn't need to be very high (3 consecutive concurrent mods would probably be reasonable enough to give up IMHO).

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Utils.java
bserdar: Yes, there is a race condition, and it is possible that a successfully
updated doc will appear as though it failed. However, technically this
would not be a false negative, because if you go and read the doc, you'd
see that someone has really updated it, just not at the moment where this
algorithm is supposed to detect. So, this is at least safe.

I couldn't find any other way to figure out what failed. Mongo doesn't
return that info. Hence the race condition.

alechenninger: That's fair. At least there is no possibility of false positive. Maybe worth adding a comment to document this known issue?

bserdar: Actually, it is not safe. In a real concurrent update error, the second
thread updates the unmodified doc, whereas in the other case, the second
thread updates the modified doc. So if two threads are both running x=x+1,
in the real concurrent update, when exception is thrown the doc will have
x+1, and in the false negative, it'll have x+2, which the client will
probably retry, and get x+3.

Let me think a bit if I can eliminate false negative...

alechenninger: Maybe if docVersions was an array? Could check if this transactions version is there. Would probably not want to keep every version number though.

alechenninger: I think you could maintain size with $slice though, so maybe that could work: https://docs.mongodb.com/manual/reference/operator/update/slice/#up._S_slice

bserdar: All solutions look half baked. Considering throwing this PR away, and
implementing transactions. Opinions?

alechenninger: What if for a request where concurrent mod is detected, we return with status "partial"? It is still preferable to make the concurrent mod detection visible to client over silently replace the effects of a previous successful update IMHO, even if the client doesn't get to see what records did not get the update. Additionally, I'd like to be able to get some reporting on how often this case actually happens.

bserdar: It'd return partial already. The problem is the updates that did not fail,
but reported as such, because they might force a retry even though the
update worked.

alechenninger: I guess I'm saying what if we don't try to report any as failed (because we don't know which), just return with a partial update. The client has to figure out what didn't update of course but, I think that is still preferable than losing previous update(s).

bserdar:
Throwing this away, will redo.

PR #300, opened 2016-08-16
Fixes #298

### LINKED ISSUE #298: Translator.populateCaseInsensitiveHiddenFields algorithm is inefficient
Below segment is extremely inefficient. It converts doc to string, parses json, and then converts that to string and parses it back to bson, and does this for every case insensitive field.

It should locate the hidden field, and add/modify the object containing the field. That way, whatever was there would also be preserved.

                    ObjectNode arrNode = JsonNodeFactory.instance.objectNode();
                    LOGGER.debug("Adding field '" + fieldMap.get(index) + "' to document with value " + dbObject.toString().toUpperCase());
                    JsonDoc.modify(arrNode, new Path(fieldMap.get(index)), JsonNodeFactory.instance.textNode(dbObject.toString().toUpperCase()), true);
                    ObjectMapper mapper = new ObjectMapper();
                    JsonNode merged = merge(mapper.readTree(doc.toString()), mapper.readTree(arrNode.toString()));
                    DBObject obj = (DBObject) JSON.parse(merged.toString());
                    ((BasicDBObject) doc).clear();
                    ((BasicDBObject) doc).putAll(obj);

BVulaj:


BVulaj:
Worth noting that we can probably make this change for array fields as well. I can either make it part of this PR or just merge this and I'll make another.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -766,30 +772,34 @@ public static void populateDocHiddenFields(DBObject doc, EntityMetadata md) thro
...
+                // given the last basic object, populate its hidden field
+                String val = (String) currentDbo.get(path.getLast());
+                if (val != null) {
+                    DBObject hidden = (DBObject) currentDbo.get(HIDDEN_SUB_PATH.toString());
+                    if (hidden == null) {
+                        currentDbo.put(HIDDEN_SUB_PATH.toString(), new BasicDBObject(path.getLast(), val.toUpperCase()));
+                    } else {
+                        hidden.put(path.getLast(), val.toUpperCase());
+                    }
                 }
```
bserdar: ...else (if val==null) and if hidden_sub_path exists, remove the field from there

bserdar:
Lets make this for arrays as well

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
BVulaj: Good catch.  We didn't have that happening before and this logic makes it quite a bit easier to implement.  I think there's another issue out there that we can close.

bserdar:
I believe you're making this much more complicated than it needs to be. First, instead of strings for field names, use Path. Second, you don't need a Map for field and its corresponding hidden field. A list of fields
(paths) should suffice. And the algorithm would look like this:

populateCaseInsensitiveField(Object doc,Path field) {
   if field.numSegments==1
       if doc.get(field) is a list (must be primitive list)
           add hidden field to doc, and populate the list
       else
           add hidden field to doc, populate field
  else if field[0]=='*'
       doc is a List. For every element of doc, populateCaseInsensitiveField(element, field.suffix(-1)
  else
       populateCaseInsensitiveField(doc.get(field[0],field.suffix(-1)
}

BVulaj:
@bserdar that's fair.  I'm still spending some time cleaning this up.  Let me take a look at the algo.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -756,162 +752,51 @@ private DBObject translateNaryValueRelationalExpression(FieldTreeNode context, N
...
-                    DBObject obj = (DBObject) JSON.parse(merged.toString());
-                    ((BasicDBObject) doc).clear();
-                    ((BasicDBObject) doc).putAll(obj);
+    public static void populateCaseInsensitiveField(Object doc, Path field) {
+        if (doc == null) {
+            return;
+        } else if (field.numSegments() == 1) {
+            DBObject docObj = (DBObject) doc;
+            if (docObj.get(field.head(0)) == null) {
+                // no value, so nothing to populate
```
bserdar: Remove field from docObj.@hidden

PR #301, opened 2016-08-18
Measure update

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -58,6 +59,7 @@
 public class IterateAndUpdate implements DocUpdater {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IterateAndUpdate.class);
+    private static final Logger METRICS = LoggerFactory.getLogger("com.redhat.lightblue.metrics");
```
paterczm: The labels are very generic. Might be a problem when we start adding measurements to other classes. I think you can leave those labels as is if you rename the logger, e.g.:
```java
LoggerFactory.getLogger("com.redhat.lightblue.metrics.IterateAndUpdate");
```

bserdar: Changed logger name to "metrics.com.redhat..."

PR #303, opened 2016-08-25
Re safe update

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+public class MongoSafeUpdateProtocol {
+
+    private static final Logger LOGGER=LoggerFactory.getLogger(MongoSafeUpdateProtocol.class);
+    
+    public static final String DOCVER="docver";
+
+    public static final String DOCVER_FLD=Translator.HIDDEN_SUB_PATH.toString()+"."+DOCVER;
+    public static final String DOCVER_FLD0=Translator.HIDDEN_SUB_PATH.toString()+"."+DOCVER+".0";
+
+    private static final long TOO_OLD=10l*60l*1000l; // Any docver older than 10 minutes is to old
```
BVulaj: Why 10 minutes?  Just a safe amount of time?

bserdar: Arbitrary time that's not too short or too long. Maybe 10 minutes is too
long? 1 minute?

alechenninger: IMHO consider naming TOO_OLD_MS (I'm a stickler for putting units in var names when it's a duration) or use JDK8 Duration type.

It sounds like our constraints on the time are this: we want to balance not having docVer array too large, but giving enough time between our update and our concurrent mod check so that it finds our current transactions docVer in the array before it is blown away. Correct? 

If so, then probably a minute or two is okay amount of time for us to find concurrent mods. It would probably depend on client's mongo and timeout configuration, though. For example, if a long update starts with MMAPv1 just after the commit, then depending on timeout configuration, maybe the collection would be locked for a while before the find concurrent updates could run, and we would miss it if another transaction came through and cleaned up the old version before our check finished.

Of course, this would be rare, cause lots of other problems (if collection locked for so long), and the side effect would be clients seeing a false positive of concurrent update (not world ending, but not good).

Maybe there is another failure condition -- when findConcurrentModifications too close to TOO_OLD? Then we cannot be certain that docs not found by check were not found because of concurrent mod, or because their docVer was removed before we could check.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+     * Commits the current batch, and prepares for the next
+     * iteration. If any update operations fail, this call will detect
+     * errors and associate them with the documents using the document
+     * index.
+     */
+    public Map<Integer,Error> commit() {
+        Map<Integer,Error> results=new HashMap<>();
+        if(!idsInBatch.isEmpty()) {
+            BulkWriteResult writeResult;
+            LOGGER.debug("attempToUpdate={}",idsInBatch.size());
```
alechenninger: Typo: attemp*t*ToUpdate

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+    /**
+     * Removes old docvers from a document
+     */
+    private void cleanupOldDocVer(DBObject doc) {
+        DBObject hidden=getHidden(doc,false);
+        if(hidden!=null) {
+            List<ObjectId> list=(List<ObjectId>)hidden.get(DOCVER);
+            if(list!=null) {
+                List<ObjectId> copy=new ArrayList<>(list.size());
+                long now=docVer.getDate().getTime();
```
alechenninger: Forgot that mongo ObjectId encodes a timestamp... nice.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+                    DBObject obj=cursor.next();
+                    Object id=obj.get("_id");
+                    successfulIds.add(id);
+                }
+                index=0;
+                for(Object id:idsInBatch) {
+                    if(!results.containsKey(index)) { // No other errors for this id
+                        if(!successfulIds.contains(id)) {
+                            // concurrency errors for this id
+                            results.put(index,Error.get("update",MongoCrudConstants.ERR_CONCURRENT_UPDATE,id.toString()));
```
alechenninger: Per above comment, is worth checking here if we may have not seen the doc because the version is too old?

Not sure what you would be able to do about it other than log.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+    private DBObject writeReplaceQuery(DBObject doc) {
+        DBObject hidden=getHidden(doc,false);
+        ObjectId top=null;
+        if(hidden!=null) {
+            List<ObjectId> list=(List<ObjectId>)hidden.get(DOCVER);
+            if(list!=null&&!list.isEmpty())
+                top=list.get(0);
+        }
+        BasicDBObject query=new BasicDBObject("_id",doc.get("_id"));
+        query.append(DOCVER_FLD0,top);
```
alechenninger: Should docver array be indexed by default?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+                for(ObjectId id:list) {
+                    if(!id.equals(docVer)) {
+                        Date d=id.getDate();
+                        if(now-d.getTime()<TOO_OLD) {
+                            copy.add(id);
+                        }
+                    } else {
+                        copy.add(id);
+                    }
+                }
```
alechenninger: Maybe we could have both timeout _and_ max array length (keeping old versions if max length not met yet). This would keep most ids possible in case it takes ridiculously long to find concurrent updates.

I might just be being paranoid on that though. Maybe it's enough to just have long enough timeout.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
bserdar: Renamed constant, and changed it to 1 min.

Your last point about not finding doc at the boundaries of TOO_OLD: someone
must have modified the doc for it to be not found, so it is still a
legitimate concurrent update error.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
bserdar: it is always used by _id, within a set of 'batchsize' docs, so index should
not be necessary.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -0,0 +1,349 @@
...
+                    LOGGER.debug("Successful update");
+                } else {
+                    LOGGER.debug("notUpdated={}",idsInBatch.size()-writeResult.getMatchedCount());
+                    findConcurrentModifications(results);
+                }
+            } catch (BulkWriteException e) {
+                List<BulkWriteError> writeErrors=e.getWriteErrors();
+                if(writeErrors!=null) {
+                    for(BulkWriteError we:writeErrors) {
+                        if (we.getCode() == 11000 || we.getCode() == 11001) {
```
BVulaj: Are there constants for these somewhere?  Worth making them constants if not?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
bserdar: The timeout is only useful for many updates taking place within that
timeout. If two threads start updating the same object, and they interleave
correctly, timeout won't matter. If they don't interleave correctly, that
means then it is already a concurrent modification error. Timeout is there
to limit the size of the array when many updates happen in a short time.

Am I thinking this correctly?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
bserdar: Done.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
alechenninger: Sorry, I added an edit, not sure if you saw by email. Mainly it would be because, if it used often, then because both field are in index, it would be a "covered query" which are very efficient: https://docs.mongodb.com/manual/indexes/#covered-queries

bserdar: Yes it would be efficient, but we're already doing an _id seach, which
should bring one document into the memory, and then it should check if that
document has the docver or not.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
BVulaj: If it's there to limit the size of the array when many updates happen in a short time, then maybe we do want a max size, rather than a timeout.  (Or both?)

alechenninger: Hrm... concurrency is always a bit mind bending :).

The scenario that could be a problem (but perhaps it is too rare depending on timeout?) I think is this:

Given timeout is 1:00 (1 minute), and there are [at least] three overlapping updates throughout this timeframe.

0. Two updates, A and B find their docs at ~0:00
1. A and B start commits at 0:01 with docvers from ~0:00
2. They finish. B is successful, returns. A misses some updates, will check for concurrent mods.
3. Update C starts at 1:00, removes some docvers from update A which are too old, and commits.
4. A finds concurrent mods at 1:01 (big delay from step 2... would this happen?), gets less back than expected because of update C removing old doc versions. It knows it updated N documents, but finds >N concurrent mods.

What do you think? Maybe to some extent we can't be perfect, and if timeout is long enough, it would be crazy to actually happen. So maybe imperfect is okay. (Assuming I don't have something wrong). Even my suggestion (keeping some old ones if array size not too big) is still imperfect (just imagine there are many successful updates in between 2 and 4 ... array size could be met eventually).

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
alechenninger: Makes sense. Turns out covered queries do not work with array fields anyways (for future reference).

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
bserdar: Yes, timeout is imperfect. The timeout is between the time A commits, and
reads it back, so if A cannot read back what it wrote within a minute, and
some other process comes and updates it, A will think it is a concurrent
update. With the existing mongo apis, the only way to deal with this is to
increase the timeout.

I think 1 minute timeout is large enough to make this scenario rare enough,
but we can increase it.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -284,6 +288,7 @@ private void updateDocs(CRUDOperationContext ctx,
                             ctx.getFactory().getInterceptors().callInterceptors(InterceptPoint.PRE_CRUD_UPDATE_DOC, ctx, doc.inputDoc);
                             merge.merge(doc.oldDoc, doc.newDoc);
                             Translator.populateDocHiddenFields(doc.newDoc, md);
+                            MongoSafeUpdateProtocol.overwriteDocVer(doc.newDoc,docver);
```
alechenninger: Maybe this should append w/ timeout clean up like other updates, to prevent removing an a doc version of a successful update that happened immediately prior to this save.

1. Update (A) partial success, needs to check for concurrent mods
2. Save (B) retrieves updates from A, but overwrites doc ver
3. Check for concurrent mods for A has false positive where overlaps with B, because docver is gone even though both are successes

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -207,58 +211,29 @@ public void update(CRUDOperationContext ctx,
             }
             measure.end("iteration");
             // if we have any remaining items to update
-            if (numUpdating > 0) {
-                try {
-                    executeAndLogBulkErrors(bwo);
-                } catch (Exception e) {
-                    LOGGER.warn("Update exception for documents for query: {}", query.toString());
+            if (docUpdateAttempts.size() > batchStartIndex) {
+                Map<Integer,Error> updateErrors=sup.commit();
```
alechenninger: Do you want to measure a "bulkUpdate" here, too?

alechenninger:
Just have concern with save update overwriting all doc versions. Otherwise LGTM.

paterczm:
If it's not too problematic, I would add a switch to be able to disable this feature. It's not a trivial change, will degrade update performance (hopefully just a little, but I'm not sure at this point) and we operate on a tight schedule.

bserdar:
It only affects update performance if there is an error. Successful updates
are not affected. I can add a flag though.

bserdar:
All changes are made. There is now a configuration option to turn this off.

Pls review and merge
https://github.com/lightblue-platform/lightblue-core/pull/685 before this
one.

alechenninger:
With updates, LGTM.

BVulaj:
Merged the lb-core PR.  The rest looks good to me.

Issue #306, opened 2016-09-13
Updates susceptible to concurrent update errors if lightblue readPreference is not primary


alechenninger:
As I understand it the algorithm is this, for an update:

1. Find the doc -- uses lightblue's configured read preference. Can read from secondary.
2. Apply lightblue update to in memory doc.
3. Write the updated doc against primary using update-if-current

It is very easy to get concurrent update errors with step 3 if lightblue is configured with non-primary read preference, because the window for concurrent update is as long as replication lag would be. So if replication lag is a second, any updates made within a second against the same document on that lightblue node will fail because the secondary does not yet have the doc version(s) of updates within that second (it is not current according to primary).

This means it's not reliable for clients to run consecutive updates, even if they do not have find in between, if they are talking to lightblue which is configured with non primary read preference.

alechenninger:
Fixed in #307

PR #307, opened 2016-09-14
Fix #305 and #306: read-for-update operations use primary

### LINKED ISSUE #305: Are lock operations done w/ primary read preference?
For example, does ping always check primary?

I looked through code, it looks like it is not set, so should default to primary (says mongo docs), but I'm not sure if that default is overridden anywhere.

I ask because I occasionally see some weird things like, a client will acquire a lock successfully, then ping it, and now they get an invalid lock exception just a half a second later, and TTL is 90 seconds.

bserdar:


alechenninger:
Is there any way to test this?

bserdar:
If this can be reproduced in dev env, we can push it, and try to reproduce
it.

alechenninger:
Sorry I mean, automated unit or integration test of some kind. I don't think it's a blocker to merge, but it might be good to have a test harness with artificially broken / slow mongo replication that we could use to test some of these weird edge cases.

bserdar:
That's pretty difficult to setup for a unit test. Let me think about it/

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -123,6 +124,8 @@ public void update(CRUDOperationContext ctx,
             measure.begin("iteration");
             int batchStartIndex=0; // docUpdateAttempts[batchStartIndex] is the first doc in this batch
             while (cursor.hasNext()) {
+                // Read from primary for read-for-update operations
+                cursor.setReadPreference(ReadPreference.primary());
```
alechenninger: I think this needs to be set before the call to hasNext -- at that point we have already executed the query right?

bserdar: fixed.

PR #312, opened 2016-09-22
Throw exception if cannot retrieve what we expect; simplify limits al…

bserdar:
…gorithm

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
@@ -77,7 +77,6 @@ public class BasicDocFinder implements DocFinder {
 
         long executionTime = System.currentTimeMillis();
         DBCursor cursor = null;
-
         try {
             cursor = coll.find(mongoQuery, mongoProjection);
             if (readPreference != null) {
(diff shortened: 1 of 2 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -97,50 +96,61 @@ public long find(CRUDOperationContext ctx,
             }
             int size = cursor.count();
```
alechenninger: cursor has a size() so it's a little confusing to me that this variable is size even though it's not the cursor size, it's the cursor count, which are two different values. So maybe count instead of size would be clearer.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -97,50 +96,61 @@ public long find(CRUDOperationContext ctx,
...
+            int f=from==null?0:from.intValue();
+            int t=to==null?size-1:to.intValue();
+            if(f<0)
+                f=0;
+            if(t<0)
+                nRetrieve=0;
+            if(t>=size)
+                t=size-1;
+            if(f>=size) 
+                nRetrieve=0;
```
alechenninger: I think the nRetrieve=0 cases are all essentially the same condition, but reworded, so maybe they could be collapsed.

- If `t < 0`, then `t` will be `< f`, because `f` is clamped at min 0. So you can just clamp `f` at 0 at then you don't need `t < 0` in addition to `t < f`.
- If `f` is `>= size`, then `t` will always be `< f`, because `t` is max size-1, so you don't need `f >= size` in addition to `t < f`.

So `t < f` covers both cases given the boundaries.

Ultimately I think this is all saying (leaving out null handling for brevity): 

1. set f = from min 0
2. set t = to max size -1
3. set nRetrieve = t-f+1 min 0

Then deal with nRetrieve accordingly as you already do. To me this algorithm is simpler to understand at a glance.

What do you think?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -97,50 +96,61 @@ public long find(CRUDOperationContext ctx,
...
-                //if 'to' is not null, check if greater than or equal to 0 and set limit 
-                if (to >= 0) {
-                    cursor.limit(to.intValue() - (from == null ? 0 : from.intValue()) + 1);
-                } // if to is less than 0, skip the entire list to return nothing
-                else if (to < 0) {
-                    cursor.skip(size);
+                
+                LOGGER.debug("Retrieving results");
+                retrievalTime = System.currentTimeMillis();
+                mongoResults = cursor.toArray();
```
alechenninger: I guess toArray() just uses the iterator as such internally, so nevermind.

bserdar:
Changes are pushed.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -95,53 +94,59 @@ public long find(CRUDOperationContext ctx,
...
+            long retrievalTime=0;
+
+            // f and t are from and to indexes, both inclusive
+            int f=from==null?0:from.intValue();
+            int t=to==null?numMatched-1:to.intValue();
+            if(f<0)
+                f=0;
+            if(t>=numMatched)
+                t=numMatched-1;
+            if(t<f||f>=numMatched)
```
alechenninger: I don't think it matters much, but I think if f >= numMatched, then t will be < f, since t is at most numMatched - 1, so `||f>=numMatched`) is maybe not necessary.

IMHO, you could consider skipping this check, and only do `int nRetrieve = Math.min(0, t - f + 1)`. Then you only need to check if nRetrieve is > 0, and it doesn't need to be reassigned again on 116.

Many ways to skin a cat, so all I really mean to say is that I think you can take out `||f>=numMatched` :-)

Other than that, LGTM

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCrudConstants.java
```diff
@@ -30,6 +30,7 @@
     public static final String ERR_NO_ACCESS = "mongo-crud:NoAccess";
     public static final String ERR_CONNECTION_ERROR = "mongo-crud:ConnectionError";
     public static final String ERR_MONGO_ERROR = "mongo-crud:DatabaseError";
+    public static final String ERR_MONGO_RESULTSET_MISMATCH = "mongo-crud:resultsetMismatch";
```
alechenninger: Maybe should be ResultSetMismatch to be consistent with other error case?

bserdar:
Fixes are pushed.

PR #315, opened 2016-09-27
313

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/ConcurrentModificationDetectionCfg.java
```diff
@@ -0,0 +1,99 @@
...
+        return failureRetryCount;
+    }
+
+    public void setFailureRetryCount(int n) {
+        failureRetryCount=n;
+    }
+
+    /**
+     * If true, updater validates that the document still matches the
+     * query before attempting a retry
```
paterczm: I would mention here that this applies only to documents which failed to update before, so it's clear what the limitation is.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -201,20 +235,20 @@ public int getSize() {
...
                 } else {
                     writeResult=bwo.execute(writeConcern);
                 }
                 LOGGER.debug("writeResult={}",writeResult);
-                if(idsInBatch.size()==writeResult.getMatchedCount()) {
+                if(batch.size()==writeResult.getMatchedCount()) {
                     LOGGER.debug("Successful update");
                 } else {
-                    LOGGER.debug("notUpdated={}",idsInBatch.size()-writeResult.getMatchedCount());
+                    LOGGER.debug("notUpdated={}",batch.size()-writeResult.getMatchedCount());
```
paterczm: Perhaps we should log this as WARN? It would be good to know how often it happens. Knowing that could help us understand the performance impact if we decide to take this further (run the query again to include new documents in the retry).

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -233,23 +267,96 @@ public int getSize() {
                 findConcurrentModifications(results);
             }
         }
+        retry(results);
```
paterczm: Really minor, but I would call this method "retryIfNeeded" or "retryIfConcurrentUpdate"

bserdar:
Tweaks pushed.

Issue #317, opened 2016-09-28
Do not call createIndex on entry to locking and sequence generator unless necessary


bserdar:
The call acquires lock

PR #318, opened 2016-09-29
Initialize locking and sequences collections only once per jvm

paterczm:
https://github.com/lightblue-platform/lightblue-mongo/issues/317

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
@@ -20,8 +20,8 @@ package com.redhat.lightblue.mongo.crud;
 
 import java.util.Arrays;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -52,7 +52,7 @@ public class MongoLocking implements Locking {
     private long defaultTTL = 60l * 60l * 1000l;// 1 hr
 
     // a set of locking collections which were already initialized
-    private static Set<String> initializedCollections = new HashSet<>();
+    private static Set<String> initializedCollections = new CopyOnWriteArraySet<>();
 
     public MongoLocking(DBCollection coll) {
         this(coll, false);
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
@@ -18,19 +18,21 @@
  */
 package com.redhat.lightblue.mongo.crud;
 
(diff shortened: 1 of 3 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
```diff
@@ -49,8 +51,24 @@
...
+
+    /**
+     *
+     * @param coll
+     * @param forceCollectionInit set to true in unit tests, where collections are wiped out between the tests
+     */
+    public MongoLocking(DBCollection coll, boolean forceCollectionInit) {
+        if (forceCollectionInit || !initializedCollections.contains(coll.getFullName())) {
+            init(coll);
+            initializedCollections.add(coll.getFullName());
```
bserdar: Collection name is not enough for the general case. Use dbname+collectionname. Also, synchronize the access to the set.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
```diff
@@ -57,6 +57,9 @@ private void initIndex() {
         coll.createIndex(keys, options);
     }
 
+    // true when sequance collection is initialized and indexes created
+    private static boolean initialized = false;
```
bserdar: There can be more than one collection used for this, so same thing used in lock should be used here.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
paterczm: Full name includes db name: http://api.mongodb.com/java/current/com/mongodb/DBCollection.html#getFullName--

paterczm: What's the worst that can happen if I don't synchronize? If it initializes a few times instead of once then it doesn't really change anything.

bserdar: Synchronize the add() call, that'll throw an exception if two threads enter it.

paterczm: How about using https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CopyOnWriteArraySet.html?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
paterczm: Fixed.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoLocking.java
bserdar: Haven't used that before. Looks ok to use that.

PR #321, opened 2016-10-06
Fix #320: If there are duplicate ids, diff computation should fallbac…

### LINKED ISSUE #320: Update diff/merge computation should handle possible duplicate array ids

bserdar:
…k to non-id comparison

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
@@ -194,12 +194,12 @@ public final class BsonMerge extends DocComparator<Object, Object, DBObject, Lis
             try {
                 return compareNodes(oldDoc,newDoc);
             } catch(DocComparator.DuplicateArrayIdentity dai) {
-                LOGGER.warn("Duplicate ID:{}");
+                LOGGER.warn("Duplicate array element id for doc._id:{}, field:{}",oldDoc.get("_id"),dai.getPath());
                 Path p=dai.getPath();
                 ignoredIdentities.add(getArray(p));
                 LOGGER.debug("Ignored paths:{}",ignoredIdentities);
             } catch(DocComparator.InvalidArrayIdentity iai) {
-                LOGGER.warn("Invalid ID:{}");
+                LOGGER.warn("Invalid array element id for doc._id:{}, field:{}",oldDoc.get("_id"),iai.getPath());
                 Path p=iai.getPath();
                 ignoredIdentities.add(getArray(p));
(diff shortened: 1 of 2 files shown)


paterczm:
Will this silently drop duplicated array elements from request? This is different behavior than at the document level, where you get constraint violation. My concern is that someone may expect the same behavior from an array.

bserdar:
This will try to merge two documents assuming the array identities are not
identities. We fixed the validation part, so this can only happen if there
is already data in the database with duplicated IDs. In that case, it will
compare array elements using content, not id.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -146,6 +148,62 @@ protected int size(List value) {
...
+                mp.push(p.head(i));
+        }
+        return mp.immutableCopy();
+    }
+
+    public DocComparator.Difference<Object> compareNodesWithFallback(DBObject oldDoc,DBObject newDoc) throws Exception {
+        // Compare two docs. If we get a duplicate array identity exception, remove that array identity from
+        // the identity map, and compare again. This is to deal with docs that contain duplicate array identities either
+        // because the identity constraint is added after data is screwed up, or becase we forgot to validate it
+        while(true) {
```
paterczm: Maybe use a for to avoid infinite loop in case there is something wrong with the loop logic? It can throw an exception if reaches 1000 or so.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -146,6 +148,62 @@ protected int size(List value) {
...
+            try {
+                return compareNodes(oldDoc,newDoc);
+            } catch(DocComparator.DuplicateArrayIdentity dai) {
+                LOGGER.debug("Duplicate ID:{}");
```
paterczm: How about a warning (log.warn) about a non unique identity in an array? This is a data integrity problem after all.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BsonMerge.java
```diff
@@ -146,6 +148,62 @@ protected int size(List value) {
...
+                Path p=dai.getPath();
+                ignoredIdentities.add(getArray(p));
+                LOGGER.debug("Ignored paths:{}",ignoredIdentities);
+            } catch(DocComparator.InvalidArrayIdentity iai) {
+                LOGGER.debug("Invalid ID:{}");
```
paterczm: Also a warning.

Issue #311, opened 2016-09-20
Add support for partial indexes


jewzaam:
https://docs.mongodb.com/manual/core/index-partial/

Beyond being a nice capability, we could leverage for unique indexes.  Today we have all unique indexes configured to be sparse, else two null index field values would be a collision.

jewzaam:
Did not create an issue for the related work necessary in core if we do this.

PR #324, opened 2016-10-20
Partial index support

paterczm:
https://github.com/lightblue-platform/lightblue-mongo/issues/311

* Using mongo 3.1.x for testing (upgrade required for partial indexing support). This means no more regression testing for 2.6.x.
* Is there any policy on backend specific option naming? Like mongo-partialFilterExpression prefix?

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
@@ -1011,17 +1011,28 @@ public class MongoCRUDController implements CRUDController, MetadataListener, Ex
         return false;
     }
 
-    private boolean indexOptionsMatch(Index index, DBObject existingIndex) {
+    /**
+     * Compare index options in metadata with existing indexes in mongo. The only 2 options
+     * that matter are 'unique' and 'partialIndexExpression'.
+     *
+     * @param index
+     * @param existingIndex
+     * @return true if all index options recognized by Lightblue match, false otherwise
+     */
+    public static boolean indexOptionsMatch(Index index, DBObject existingIndex) {
         Boolean unique = (Boolean) existingIndex.get("unique");
-        if (unique != null) {
-            if ((unique && index.isUnique())
-                    || (!unique && !index.isUnique())) {
-                return true;
-            }
-        } else if (!index.isUnique()) {
-            return true;
+
+        if (!new Boolean(index.isUnique()).equals(unique)) {
+            LOGGER.debug("Index unique flag changed to {}",unique);
+            return false;
         }
-        return false;
+
+        if (index.getProperties().containsKey(PARTIAL_FILTER_EXPRESSION_OPTION_NAME)) {
+            LOGGER.debug("Index partialFilterExpression option changed from {} to {}", existingIndex.get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME), index.getProperties().get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME));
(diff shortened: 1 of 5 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -815,12 +818,17 @@ private void createUpdateEntityInfoIndexes(EntityInfo ei, Metadata md) {
...
-                options.append("sparse", index.isUnique());
+                // if index is unique and non-partial, also make it a sparse index, so we can have non-required unique fields
+                options.append("sparse", index.isUnique() && !index.getProperties().containsKey(PARTIAL_FILTER_EXPRESSION_OPTION_NAME));
                 if (index.getName() != null && index.getName().trim().length() > 0) {
                     options.append("name", index.getName().trim());
                 }
                 options.append("background", true);
+                // partial index
+                if (index.getProperties().containsKey(PARTIAL_FILTER_EXPRESSION_OPTION_NAME)) {
+                    DBObject filter = (DBObject) JSON.parse(index.getProperties().get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME).toString());
```
bserdar: iirc, this is not how the metadataparser will parse this. If the property is:

{ field: { .. } }

then, the property will return a Map, with a key "field", and value of another map, and so on.

You need to recursively process this and build a BasicDBObject from that, instead of using json parser. That could be useful, so you can add that as a static method to PropertyParser in core.

paterczm: Modified the test to actually parse the partialFilterExpression property and it still works. Am I missing something?

bserdar: The property is  "partialFilterExpression": "{ \"field3\": { \"$gt\": 5 } }", which is a string.

I'd much rather this be: "partialFilterExpression": { "field3": { "$gt": 5 } }

When you do this, the property value will be a map containing field3 key, with value being a map containing $gt as key, and 5 as value.

paterczm: For readability's sake it makes sense.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -998,17 +1011,28 @@ protected boolean indexFieldsMatch(Index index, DBObject existingIndex) {
...
+
+        if (!new Boolean(index.isUnique()).equals(unique)) {
+            LOGGER.debug("Index unique flag changed to {}",unique);
+            return false;
         }
-        return false;
+
+        if (index.getProperties().containsKey(PARTIAL_FILTER_EXPRESSION_OPTION_NAME)) {
+            LOGGER.debug("Index partialFilterExpression option changed from {} to {}", existingIndex.get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME), index.getProperties().get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME));
+            return index.getProperties().get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME).equals(existingIndex.get(PARTIAL_FILTER_EXPRESSION_OPTION_NAME));
```
bserdar: My guess is this comparison won't work when existingIndex is loaded from the DB. Index.getProperties().get() will have an object tree containing Maps and Lists, existingIndex.get() will have an object tree containing DBObjects and DBLists.

In the unit test, you should test this with an existingIndex retrieved from the DB.

PR #330, opened 2016-10-28
Fixing logic to identify index unique option changes

paterczm:
I introduced a regression with https://github.com/lightblue-platform/lightblue-mongo/pull/324.


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -1022,8 +1022,9 @@ protected boolean indexFieldsMatch(Index index, DBObject existingIndex) {
     public static boolean indexOptionsMatch(Index index, DBObject existingIndex) {
         Boolean unique = (Boolean) existingIndex.get("unique");
 
-        if (!new Boolean(index.isUnique()).equals(unique)) {
-            LOGGER.debug("Index unique flag changed to {}",unique);
+        // existing unique can be null, that's the same as false
```
bserdar: This is hard to read. How about:

if( (unique!=null&&unique)!=index.isUnique() )

paterczm: I would argue that this is more readable. Made a different change, let me know what you think.

Issue #329, opened 2016-10-28
Cannot specify partialFilterExpression for a field in an array or subdocument


paterczm:
because Mongo forbids field names with a dot [1]. A partialFilterExpression like `{"authentication.prividerName": "Red Hat"}` will result with a "Invalid BSON field name" when saving metadata:

```
2016-10-28 19:26:05,647 ERROR [com.redhat.lightblue.mongo.metadata.MongoMetadata] (default task-9) updateEntityInfo: java.lang.IllegalArgumentException: Invalid BSON field name authentications.providerName
        (...)
        at com.redhat.lightblue.mongo.metadata.MongoMetadata.updateEntityInfo(MongoMetadata.java:362)
```

[1] https://docs.mongodb.com/manual/core/document/#field-names

bserdar:
Instead of converting this property to BSON, we should save it as a string of JSON doc.

paterczm:
Perhaps replace dots in field names with "$dot$", in a way transparent to the user?

PR #331, opened 2016-10-31
Define partialFilterExpression as a string after all

paterczm:
to be able to define fields with a dot in it's name: https://github.com/lightblue-platform/lightblue-mongo/issues/329

This is not ideal (json defined as a value in json), but I'm in a hurry before today's release.


bserdar:
There is a better way of doing this, I believe. 
  - Leave the metadata expression unescaped,
  - Change the metadata translation to save that expression as a string, and load it back as a json
That leaves the CRUDController code and tests unchanged.

bserdar:
Changing metadata translation can be done in a few ways, one way is to override parseIndex/convertIndexes, and modify the property in BSON with post/pre-processing.

paterczm:
Since this is a limitation in mongo persistence, shouldn't the fix belong to lightblue-mongo?

bserdar:
It is in lightblue-mongo. The class that translates metadata to/from BSON is mongo/.../metadata/BSONParser

paterczm:
Obsoleted by https://github.com/lightblue-platform/lightblue-mongo/pull/332.

PR #332, opened 2016-10-31
Transparently convert partialFilterExpression index property to string

paterczm:
and back to be able to persist expressions referencing subdocuments or arrays.

https://github.com/lightblue-platform/lightblue-mongo/issues/329

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
@@ -19,9 +19,8 @@
 package com.redhat.lightblue.mongo.metadata;
 
 import java.util.ArrayList;
-import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import org.bson.BSONObject;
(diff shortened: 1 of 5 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -132,14 +126,52 @@ public Object asRepresentation(Object value) {
...
+                if (partialFilterExpression != null) {
+                    // convert string to json
+                    // https://github.com/lightblue-platform/lightblue-mongo/issues/329
+                    index.put(MongoCRUDController.PARTIAL_FILTER_EXPRESSION_OPTION_NAME, (Map<String,Object>)JSON.parse(partialFilterExpression));
+                }
+            }
+        }
+
+        return super.parseEntityInfo(object);
+    }
```
bserdar: Ideally, you should convert it back to string after you're done with it, in case the BSON object is used somewhere else. It is invalid at this point.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -132,14 +126,52 @@ public Object asRepresentation(Object value) {
     }
 
     /**
-     * Override to set _id appropriately.
+     * Override to convert partialFilterExpression index property from string to map.
+     *
+     */
+    @Override
+    public EntityInfo parseEntityInfo(Object object) {
```
bserdar: You can simplify this by overriding parseIndex. Then it'll be called for each index.

mongo/src/main/java/com/redhat/lightblue/mongo/metadata/BSONParser.java
```diff
@@ -132,14 +126,52 @@ public Object asRepresentation(Object value) {
...
+        }
+
+        return super.parseEntityInfo(object);
+    }
+
+    /**
+     * Override to set _id appropriately and convert partialFilterExpression index property to string.
      */
     @Override
     public BSONObject convert(EntityInfo info) {
```
bserdar: You can override convert(Index) or convert(Indexes).

PR #337, opened 2016-11-21
Datecmp

bserdar:
JS Date comparisons must use date.valueOf()

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/JSQueryTranslator.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/JSQueryTranslator.java
@@ -195,22 +195,32 @@ public class JSQueryTranslator {
             // Both fields are arrays
             if(query.getOp()==BinaryComparisonOperator._neq) {
                 parentBlock.add(new SimpleStatement("%s=true",comparisonBlock.resultVar));
+                SimpleExpression cmp=new SimpleExpression(((ArrayField)rFieldMd).getElement().getType() instanceof DateType &&
+                                                          ((ArrayField)lFieldMd).getElement().getType() instanceof DateType?
+                                                          "this.%s[%s].valueOf()!=this.%s[%s].valueOf()":
+                                                          "this.%s[%s]!=this.%s[%s]",
+                                                          lfieldName,loopVar,
+                                                          rfieldName,loopVar);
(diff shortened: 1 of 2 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/JSQueryTranslator.java
```diff
@@ -219,20 +229,25 @@ private Block translateFieldComparison(Context ctx,FieldComparisonExpression que
...
             if(rFieldMd instanceof ArrayField) {
                 arrayFieldName=rfieldName;
                 simpleFieldName=lfieldName;
                 op=query.getOp().invert();
+                isDate=((ArrayField)rFieldMd).getElement().getType() instanceof DateType && lFieldMd.getType() instanceof DateType;
             } else {
                 arrayFieldName=lfieldName;
                 simpleFieldName=rfieldName;
                 op=query.getOp();
+                isDate=((ArrayField)lFieldMd).getElement().getType() instanceof DateType && rFieldMd.getType() instanceof DateType;
```
dcrissman: Didn't we just confirm that lFieldMd is NOT an ArrayField?

dcrissman:
@bserdar - Just for reference purposes, could you speak to why we need this?

Also I don't understand the change to the unit test. Is there an existing test case for this?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/JSQueryTranslator.java
bserdar: At this point, only one of lField or rField is an ArrayField. The if statement is for rField being an array, and this is in the else, so at this point, lField is an array.

bserdar:
This particular PR is to deal with Date comparison in Javascript code. It
turns out if you compare date values using <, <=, >, >=, it compares date
values, but for == and !=, it compares date objects, so if the field is a
date, we compare field.valueOf() instead of just field.

The change in the unit test simply removes one of the temp fields that we
no longer generate.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/JSQueryTranslator.java
dcrissman: Gotcha. I missed the nuance of the 'l' and 'r' prefixes.

bserdar:
'l' is 'left'
'r' is 'right'

PR #338, opened 2016-11-24
Fix #335: Add new variable resolution algorithm that knows how to han…

### LINKED ISSUE #335: Translator cannot deal with $this and $parent

bserdar:
…dle this and parent

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
@@ -195,30 +195,6 @@ public class Translator {
     }
 
     /**
-     * Translate a path to a javascript path
-     *
-     * Path cannot have *. Indexes are put into brackets
-     */
-    public static String translateJsPath(Path p) {
-        StringBuilder str = new StringBuilder();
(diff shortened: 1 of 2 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java
```diff
@@ -660,6 +636,75 @@ private FieldTreeNode resolve(FieldTreeNode context, Path field) {
...
+            while(trc!=null) {
+                if(trc==context) {
+                    fieldIsUnderContext=true;
+                    break;
+                } else {
+                    trc=trc.getParent();
+                }
+            }
+            if(!fieldIsUnderContext)
+                throw new NeedsJS();
```
dcrissman: Would it make more sense to determine if fieldIsUnderContext before calculating p and s? In the case that javascript is needed it would be better to fail fast.

bserdar:
I thought the same thing, but then realized that $parent.child.x is not the
same thing as 'x', because once you move a level up from the current
element, then there is no current element. So, even thought $parent.child.x
is under the same context, it needs to be handled with js.

PR #344, opened 2017-01-26
Update if same

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/DocVerUtil.java
```diff
@@ -0,0 +1,119 @@
...
+import com.mongodb.DBObject;
+import com.mongodb.BasicDBObject;
+
+import org.bson.types.ObjectId;
+
+public class DocVerUtil {
+
+    public static final String DOCVER="docver";
+
+    public static final long TOO_OLD_MS=1l*60l*1000l; // Any docver older than 1 minute is to old
```
paterczm: Wondering how this works. Docver is generated when a document is written, right?

mongo/src/test/java/com/redhat/lightblue/mongo/crud/UpdateIfSameTest.java
```diff
@@ -0,0 +1,227 @@
...
+        doc.put("field","updated"+doc.get("_id").toString());
+        
+        doc=getDoc("3","3",new ObjectId());
+        ver=DocIdVersion.getDocumentVersion(doc);
+        coll.insert(doc);
+        updater.addVersion(ver);
+        updater.addDoc(doc);
+        doc.put("field","updated"+doc.get("_id").toString());
+
+        // Id:4 will not be updated, its version is unknown
```
paterczm: In what circumstances can the version be unknown?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/DocVerUtil.java
bserdar: docver is initilized in MongoSafeUpdateProtocol and UpdateIfSameProtocol, when you instantiate the class. All updates written by those two set the version in the document.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/UpdateIfSameTest.java
bserdar: I don't understand the question. Version is always known, it is either null, or the value at the top of the hidden array in the doc.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/DocVerUtil.java
paterczm: I don't get it. A docver on a document (in mongo) needs to exist prior to update, so it can be compared with docver sent with the request.

paterczm: Can I do this, as Lightblue client?
1. Read entity
2. Make changes
3. Save entity back to Lightblue, knowing that it will err if the entity was modified in the mean time (it's docver changed).

Considering this, when is the docver created initially?

bserdar: There is a docver in the document, it is kept in a hidden field:
{
  @mongoHidden: {
       docver: [ v1, v2... ]
   }
}

bserdar:
There are two ways of doing that:
 1) Read the entity from lb. That also returns ResultMetadata for each doc.
ResultMetadata contains the version. When you update, set the same
ResultMetadata in the request, and set updateIfCurrentFlag.

 2) Entity metadata inject docver in the document. When you read the entity
from lb, the document contains the docver. Change the doc, and submit the
document with the same docver.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/DocVerUtil.java
paterczm: And it's saved on every write, correct?

bserdar:
Correct.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/UpdateIfSameTest.java
paterczm: Null is what I meant by unknown. I imagine that documents created prior to this change will not have a docver. What happens when an update request with update-if-current: true comes and docver in mongo is null?

bserdar:
Null is a valid version number. It will be changed to non-null with the
first update. The version itself will come back as null, and if there is a
non-null version in the db, it means it was changed.

paterczm:
This is failing after refactoring in core, but otherwise looks good.

paterczm:
Actually, wondering about one more scenario. An update with updateIfCurrent: false will generate a new docver for affected documents, right?

bserdar:
A new docver is generated every time, regardless of those flags.

PR #347, opened 2017-02-21
Add streaming result sets support

bserdar:
Stream result sets instead of storing them in lists


mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -172,7 +175,6 @@ public void update(CRUDOperationContext ctx,
             cursor.setReadPreference(ReadPreference.primary());
             measure.end("collection.find");
             LOGGER.debug("Found {} documents", cursor.count());
-            ctx.getFactory().getInterceptors().callInterceptors(InterceptPoint.POST_CRUD_UPDATE_RESULTSET, ctx);
```
bvulaj: Keeping pre_crud but ditching post_crud?

bserdar:
Yes, post_* screws everything up, and it is not used anyway. With post_*, I
can't stream, because post_* interceptors need access to the whole
resultset.

paterczm:
Isn't post update hook used for integration? @alechenninger, you'll know.

bserdar:
This doesn't change the hook processing. All hooks are post_ hooks.

paterczm:
Hooks != interceptors. So never mind.

PR #350, opened 2017-03-06
Fix #349: JS generation should check the validity of all levels in a …

### LINKED ISSUE #349: JS generation: need to check validity of multi-level fields
JS generation uses if(typeof field != undefined) before accessing a field, but if field is multilevel, then it fails if the first level isn't defined, for instance if(typeof x.y != undefined) fails if x isn't defined.

bserdar:
…field name

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/IfStatement.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/IfStatement.java
@@ -28,9 +28,6 @@ public class IfStatement extends Block {
     }
 
     public static IfStatement ifDefined(Name var,Statement...s) {
-        StringBuffer b=new StringBuffer(64);
-        for(int i=0;i<var.length();i++) {
-        }
         return new IfStatement(new SimpleExpression(buildNotUndefined(var)),s);
     }
     
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/IfStatement.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/IfStatement.java
@@ -28,11 +28,27 @@ public class IfStatement extends Block {
     }
 
     public static IfStatement ifDefined(Name var,Statement...s) {
-        return new IfStatement(new SimpleExpression("typeof this.%s != 'undefined'", var),s);
+        StringBuffer b=new StringBuffer(64);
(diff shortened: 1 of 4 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/js/IfStatement.java
```diff
@@ -28,11 +28,27 @@ public IfStatement(Expression test,Statement...s) {
     }
 
     public static IfStatement ifDefined(Name var,Statement...s) {
-        return new IfStatement(new SimpleExpression("typeof this.%s != 'undefined'", var),s);
+        StringBuffer b=new StringBuffer(64);
+        for(int i=0;i<var.length();i++) {
+        }
```
paterczm: What is this empty loop for?

bserdar:
Leftover from refactoring. Removed.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/TranslatorTest.java
```diff
@@ -326,11 +326,11 @@ public void translateNullReference() throws Exception {
     @Test
     public void translateJS() throws Exception {
         DBObject obj = expressionTranslator.translate(md, query("{'field':'field7.*.elemf1','op':'=','rfield':'field7.*.elemf2'}"));
-        Assert.assertEquals("function() {var r0=false;{if(typeof this.field7 != 'undefined'){for(var ri1=0;ri1<this.field7.length;ri1++){if(typeof this.field7 != 'undefined'){for(var ri2=0;ri2<this.field7.length;ri2++){r0=field7[ri1].elemf1 == field7[ri2].elemf2;if(r0){break;}}}if(r0){break;}}}}return r0;}",
+        Assert.assertEquals("function() {var r0=false;{if(typeof this.field7!='undefined'){for(var ri1=0;ri1<this.field7.length;ri1++){if(typeof this.field7!='undefined'){for(var ri2=0;ri2<this.field7.length;ri2++){r0=field7[ri1].elemf1 == field7[ri2].elemf2;if(r0){break;}}}if(r0){break;}}}}return r0;}",
```
dcrissman: Is it worth touching this class just for spacing changes?

bserdar:
The test compares with spaces, so I had to fix the string to match the
generated string.

PR #351, opened 2017-03-22
Close unused cursor

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -134,6 +136,9 @@ public long find(CRUDOperationContext ctx,
             }            
             return numMatched;
         } finally {
+            if(cursor!=null&&!cursorInUse) {
```
dcrissman: Why would a cursor still be in use at this point?

bserdar: line 125: cursor is returned in CursorStream. In that case it should not be closed,

PR #360, opened 2017-05-13
Fix #359: merge adds references of elements from the old doc into the…

### LINKED ISSUE #359: Concurrent update retry update will never match a document
We are seeing that when a concurrent update is detected, retries happen, but these retries never work, even if there is no other concurrent writes going on. We have reproduced this in tests, so we know there are no other concurrent writes.

I think the issue is here:

```
                    // Update the doc ver to our doc ver. This doc is here
                    // because its docVer is not set to our docver, so
                    // this is ok
                    DocVerUtil.setDocVer(newDoc,docVer);
                    // Using bulkwrite here with one doc to use the
                    // findAndReplace API, which is lacking in
                    // DBCollection
                    BulkWriteOperation nestedBwo=collection.initializeUnorderedBulkOperation();
                    nestedBwo.find(writeReplaceQuery(updatedDoc)).replaceOne(newDoc);
```

This writes the replace query after we have set the new doc version. This new query will never match, because that new doc version is of course not going to be the current; it's the new one in memory.

So writeReplaceQuery probably needs to run before setDocVer.

bserdar:
… new doc, and one of these references is the version info, so any modifications to the version info after the merge modifies both docs

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
@@ -272,7 +272,15 @@ public abstract class MongoSafeUpdateProtocol implements BatchUpdate {
             if(updatedDoc!=null) {
                 // if updatedDoc is null, doc is lost. Error remains
                 DBObject newDoc=reapplyChanges(index,updatedDoc);
+                // WARNING There is a potential for major screwup
+                // here, so be careful when you're changing this
+                // code. reapplyChanges implementation calls merge(),
+                // which adds a reference to @mongoHidden at the root
+                // level in the old document into the new
+                // document. This is NOT a copy, a reference. So, the
+                // setDocVer call below sets the version of BOTH docs
                 if(newDoc!=null) {
+                    DBObject replaceQuery=writeReplaceQuery(updatedDoc);
                     // Update the doc ver to our doc ver. This doc is here
                     // because its docVer is not set to our docver, so
                     // this is ok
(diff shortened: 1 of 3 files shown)


mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocolTest.java
```diff
@@ -223,7 +223,16 @@ protected DBObject reapplyChanges(int docIndex,DBObject doc) {
                     System.out.println("Retrying");
                     BasicDBObject newDoc=new BasicDBObject();
                     newDoc.putAll(doc);
-                    newDoc.removeField("@mongoHidden");
+                    // Emulating merging behavior here. When we merge
+                    // documents, we put references to missing items
+                    // from the old document into the new
+                    // document. Once of these references is
```
dcrissman: Typo: Once = One?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -272,7 +272,15 @@ public void retryConcurrentUpdateErrorsIfNeeded(Map<Integer,Error> results) {
             if(updatedDoc!=null) {
                 // if updatedDoc is null, doc is lost. Error remains
                 DBObject newDoc=reapplyChanges(index,updatedDoc);
+                // WARNING There is a potential for major screwup
+                // here, so be careful when you're changing this
+                // code. reapplyChanges implementation calls merge(),
+                // which adds a reference to @mongoHidden at the root
+                // level in the old document into the new
+                // document. This is NOT a copy, a reference. So, the
+                // setDocVer call below sets the version of BOTH docs
```
dcrissman: It might be worth stating why a reference is preferable to a copy.

bserdar:
Actually, it is not preferable. A deep copy is preferable. Let me think
about this a bit.

bserdar:
This is now updated.

dcrissman:
@bserdar - I don't see that any new changes were pushed up. Am I missing something?

bserdar:
No, I pushed it to wrong branch. It is there now.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
dcrissman: Does this comment need to be updated now?

dcrissman:
I am good with this change now. I just had one question and there seems to be a merge conflict.

bserdar:
both fixed.

PR #363, opened 2017-06-07
WIP: Changes for Monitoring and Alerting (health check using dropwizard-metrics)

sameer31:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -98,1014 +100,1053 @@
...
-            }
-        } else if (msg == null) {
-            return Error.get(otherwise, e.getMessage());
-        } else if (specialHandling) {
-            return Error.get(otherwise, msg, e);
-        } else {
-            return Error.get(otherwise, msg);
-        }
-    }
+	public static final String ID_STR = "_id";
```
bserdar: Please avoid reformatting large sections of code.

sameer31: @bserdar, will be more carefull in future.
@alechenninger, @derek63, @svkm102, Kindly review the changes

mongo/src/test/java/com/redhat/lightblue/mongo/crud/GeneratorSupportTest.java
```diff
@@ -57,6 +59,11 @@ public DB get(MongoDataStore store) {
             public MongoConfiguration getConfiguration(MongoDataStore store) {
                 return null;
             }
+
+			@Override
```
macgregor: Whats the lightblue project standard on tabs v space? We should make sure everyone is doing the same thing, whatever that is.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
alechenninger: Please undo the formatting changes so we can review. It also makes it hard to understand who modified a piece of code, because if we git blame on this file it will say all you because you reformatted, but you didn't really modify the logic. 

It looks like you are using tabs but the code in this project uses spaces. I would keep using spaces instead of tabs (pretty much all of our projects use spaces instead of tabs).

mongo/src/main/java/com/redhat/lightblue/mongo/config/MongoDBResolver.java
```diff
@@ -55,6 +58,18 @@ public MongoConfiguration getConfiguration(MongoDataStore store) {
...
+    @Override
+    public List<MongoConfiguration> getConfigurations() {
+    	
+    	List<MongoConfiguration> mongoConfigurations = new ArrayList<>();
+        
+    	for (MongoConfiguration cfg : datasources.values()) {
+    		mongoConfigurations.add(cfg);
+        }
+    	
+        return Collections.unmodifiableList(mongoConfigurations);
```
alechenninger: I think it's okay to avoid copying the list here and just `return Collections.unmodifiableList(datasources.values())`.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -104,7 +106,21 @@ public DB get(MongoDataStore store) {
...
+
+			@Override
+			public List<MongoConfiguration> getConfigurations() {
+				MongoConfiguration configuration = new MongoConfiguration();
+				try {
+					configuration.addServerAddress(db.getMongo().getAddress().getHost(), db.getMongo().getAddress().getPort());
+					configuration.setDatabase("mongo");
+				} catch (UnknownHostException e) {
+					// Exception
+				}
```
alechenninger: I'd add this to getConfiguration probably, and reference it in this method.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -2851,4 +2867,10 @@ public void indexFieldsMatch_with_lightblue_path() throws Exception {
         DBObject existingIndex = new BasicDBObject("key", new BasicDBObject("a.b", 1));
         Assert.assertTrue(controller.indexFieldsMatch(index, existingIndex));
     }
+    
+    @Test
+    public void checkHealth() throws Exception{
+    	LightblueHealth healthCheck = controller.checkHealth();
+    	Assert.assertTrue(healthCheck.isHealthy());
+    }
```
alechenninger: How can we test that when mongo is unhealthy, the health check indicates that?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -1108,4 +1107,29 @@ private Error analyzeException(Exception e, final String otherwise, final String
...
+        Collection<MongoConfiguration> configs = dbResolver.getConfigurations();
+        List<String> details = new ArrayList<>(configs.size());
+        DBObject ping = new BasicDBObject("ping", 1);
+
+        for (MongoConfiguration config : configs) {
+            try {
+                CommandResult result = config.getDB().command(ping);
+                if (!result.get("ok").equals(1.0)) {
+                    isHealthy = false;
+                    details.add(new StringBuilder(config.toString()).append("\n").append("ping:NOK").toString());
```
svkm102: ping:NOT_OK for would be better instead of ping:NOK

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -2851,4 +2870,11 @@ public void indexFieldsMatch_with_lightblue_path() throws Exception {
         DBObject existingIndex = new BasicDBObject("key", new BasicDBObject("a.b", 1));
         Assert.assertTrue(controller.indexFieldsMatch(index, existingIndex));
     }
+    
+    @Test
+    public void checkHealth() throws Exception{
+    	CRUDHealth healthCheck = controller.checkHealth();
+    	System.out.println("CRUDHealthCheck Details : " + healthCheck.details());
```
svkm102: Remove sysout

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -1,18 +1,14 @@
...
  This program is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
  the Free Software Foundation, either version 3 of the License, or
  (at your option) any later version.
-
  This program is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  GNU General Public License for more details.
-
```
alechenninger: Why the whitespace removals?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
alechenninger: Agree with Sunny. Also, newlines aren't rendered in JSON strings, so it might be better to use a different separator than a newline for readability, maybe : or = or ->, something like that.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -2851,4 +2870,11 @@ public void indexFieldsMatch_with_lightblue_path() throws Exception {
...
+    	Assert.assertTrue(healthCheck.isHealthy());
+    }
```
alechenninger: How can we test the check detects when mongo is not healthy?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
sameer31: Well, I did this because I planned to add MongoConfiguration details in the health check details which has following pattern in toString implementation (an excerpt) -
append("database:").append(database).append('\n').
                append("ssl:").append(ssl).append('\n')

However, I would change it use "=>" to display more elegantly in JSON and would not display the entire config details but just the mongo server details and the database name.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
sameer31: Adding a test case where the controller is initialized with a DB Resolver with wrong port (one other than on which in memory mongo is running) which will result in refused connection and in turn an unhealthy connection. Let me know if this would be OK. can't think of any other way right now.

mongo/src/test/java/com/redhat/lightblue/mongo/crud/MongoCRUDControllerTest.java
```diff
@@ -2851,4 +2870,49 @@ public void indexFieldsMatch_with_lightblue_path() throws Exception {
...
+
+                List<MongoConfiguration> configs = new ArrayList<>();
+                configs.add(getConfiguration(null));
+                return configs;
+            }
+        });
+        
+        CRUDHealth healthCheck = unhealthyController.checkHealth();
+        Assert.assertFalse(healthCheck.isHealthy());
+    }
```
alechenninger: Nice!

alechenninger:
This looks good! Please merge all of the updates together so there are no compile errors. I think you should be able to merge lightblue-rest last so we can still work on that.

Issue #362, opened 2017-06-06
Used pooled sequences for sequence generation


bserdar:
In metadata, add poolSize:n where n>0. When sequence is generated, add n to the sequence, and use the generated numbers for the next n calls.

bserdar:
https://github.com/lightblue-platform/lightblue-mongo/pull/367

PR #367, opened 2017-06-16
WIP: pooled sequence generator--needs threading tests, needs metadata…

bserdar:
… tests

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
@@ -18,8 +18,11 @@
  */
 package com.redhat.lightblue.mongo.crud;
 
-import java.util.Set;
+import java.util.Map;
+import java.util.HashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


dcrissman:
What is a "pooled sequence generator"?

bserdar:
Generates a pool of ids from the DB, serves the ids from the pool until all
the ids in the pool are consumed, then generates another pool of ids.
Lowers traffic on the sequence generation collection by 1/n, n being
poolsize.

bserdar:
https://github.com/lightblue-platform/lightblue-mongo/issues/362

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
```diff
@@ -80,48 +109,96 @@ private void initIndex() {
...
-        if (doc == null) {
-            // Sequence document does not exist. Insert a new document using the init and inc
-            LOGGER.debug("inserting sequence record name={}, init={}, inc={}", name, init, inc);
-            if (inc == 0) {
-                inc = 1;
+        // First check if there is already a pool of ids available
+        String fullName=coll.getFullName()+"."+name;
+        rwl.readLock().lock();
+        SequenceInfo si=sequenceInfo.get(fullName);
+        rwl.readLock().unlock();
```
dcrissman: I am not sure that I understand the lock/unlock around Hash.get. What does that achieve?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
```diff
@@ -80,48 +109,96 @@ private void initIndex() {
...
+        if(si==null) {
+            rwl.writeLock().lock();
+            si=sequenceInfo.get(fullName);
+            if(si==null) {
+                si=new SequenceInfo(fullName);
+                sequenceInfo.put(fullName,si);
             }
+            rwl.writeLock().unlock();
```
dcrissman: Looks more or less like a lazy instantiation. Would a synchronised method that returns a ReentrantReadWriteLock help clear some of this up?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
```diff
@@ -80,48 +109,96 @@ private void initIndex() {
...
+        }
 
-            BasicDBObject u = new BasicDBObject().
+        si.lock.lock();
+
+        long ret=0;
+
+        try {
+            // If there are ids in the pool, use one
+            if(si!=null) {
```
dcrissman: Is it even possible for si to be null at this point?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
bserdar: Readers of the hashmap use readLock, writers use writeLock. So, many can read, only one can write. Readers will block when someone is writing, but readers won't block each other. Better than synchronized.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
bserdar: It is lazy initialization of a pool for the first time in the hashmap. To write into the hashmap, we need a writeLock. I don't understand your comment about the synchronized method.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
bserdar: it is not, at this point, i'll clean that up

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
dcrissman: Thanks for the explanation.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSequenceGenerator.java
dcrissman: Disregard then, based on other conversation thread.

bserdar:
This is now ready for review-merge

PR #370, opened 2017-06-22
If streaming, don't compute result set size

bserdar:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -85,7 +85,7 @@ public long find(CRUDOperationContext ctx,
                 cursor.setReadPreference(readPreference);
             }
 
-            if (maxQueryTimeMS > 0) {
+            if (maxQueryTimeMS > 0&&ctx.isLimitQueryTime()) {
```
paterczm: Why this new isLimitQueryTime flag?

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
```diff
@@ -96,33 +96,48 @@ public long find(CRUDOperationContext ctx,
                 cursor = cursor.sort(mongoSort);
                 LOGGER.debug("Result set sorted");
             }
-            int numMatched = cursor.count();
-            int nRetrieve=numMatched;
+
+            int numMatched=0;
+            int nRetrieve=0;
+            if(ctx.isComputeMatchCount()) {
```
paterczm: Perhaps we can rename this flag to computeCounts and use it for modifiedCount too? https://github.com/lightblue-platform/lightblue-core/issues/789

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
bserdar: I added that thinking with streaming, there is no need, but that will still be useful if you submit a long running query, so it is there just in case.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
bserdar: i'll rename

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocFinder.java
paterczm: Still don't know why maxQueryTimeMS is not enough in streaming context.

bserdar:
It is enough. What I wanted to do is to not limit query times for
streaming, but then I have second thoughts about it. So, we're still
limiting long running queries, but we may consider disabling that for
streaming finds.

PR #376, opened 2017-07-13
Fix #374: Keep track of documents lost during update, and don't inclu…

### LINKED ISSUE #374: Concurrent updates that no longer match query are still included in response
It looks from lightblue like the update was successful even though it wasn't.

This completely breaks update-if-current algorithms.

bserdar:
…de them in the results

### DIFF
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
@@ -233,7 +233,7 @@ public class BasicDocSaver implements DocSaver {
         }
         LOGGER.debug("Save docs={}, insert docs={}, error docs={}", saveList.size(), insertList.size(), batch.size() - saveList.size() - insertList.size());
         insertDocs(ctx, collection, insertList);
-        updateDocs(ctx, collection, saveList);
+        updateDocs(ctx, collection, saveList, upsert);
     }
 
     private void insertDocs(CRUDOperationContext ctx,
@@ -315,10 +315,11 @@ public class BasicDocSaver implements DocSaver {
                                                       updateAttemptList);
         }
     }
-
+    
     private void updateDocs(CRUDOperationContext ctx,
                             DBCollection collection,
-                            List<DocInfo> list) {
+                            List<DocInfo> list,
+                            boolean upsert) {
         if (!list.isEmpty()) {
             LOGGER.debug("Updating {} docs", list.size());
             if (!md.getAccess().getUpdate().hasAccess(ctx.getCallerRoles())) {
@@ -362,8 +363,21 @@ public class BasicDocSaver implements DocSaver {
                         for(Map.Entry<Integer,Error> entry:ci.errors.entrySet()) {
                             updateAttemptList.get(entry.getKey()).inputDoc.addError(entry.getValue());
                         }
-                        for(Integer i:ci.lostDocs) {
-                            updateAttemptList.get(i).inputDoc.addError(Error.get("update",MongoCrudConstants.ERR_DOC_NO_LONGER_AVAILABLE,""));
+                        // If there are docs that were read, but then removed from the db before we updated them:
+                        //   If upsert=false, error
+                        //   If upsert=true, try reinserting them
+                        if(!upsert) {
+                            for(Integer i:ci.lostDocs) {
+                                updateAttemptList.get(i).inputDoc.addError(Error.get("update",MongoCrudConstants.ERR_SAVE_ERROR_INS_WITH_NO_UPSERT,""));
+
+                            }
+                        } else {
+                            // Try inserting these docs
+                            List<DocInfo> insList=new ArrayList<>(ci.lostDocs.size());
+                            for(Integer i:ci.lostDocs) {
+                                insList.add(updateAttemptList.get(i));
+                            }
+                            insertDocs(ctx,collection,insList);
                         }
                     } catch (RuntimeException e) {
                     } finally {
--- a/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
+++ b/mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
@@ -358,10 +358,13 @@ public class BasicDocSaver implements DocSaver {
                         doc.inputDoc.setCRUDOperationPerformed(CRUDOperation.UPDATE);
                     }
                     try {
-                        Map<Integer,Error> errorMap=upd.commit();
-                        for(Map.Entry<Integer,Error> entry:errorMap.entrySet()) {
+                        BatchUpdate.CommitInfo ci=upd.commit();
+                        for(Map.Entry<Integer,Error> entry:ci.errors.entrySet()) {
                             updateAttemptList.get(entry.getKey()).inputDoc.addError(entry.getValue());
-                        }                        
+                        }
+                        for(Integer i:ci.lostDocs) {
+                            updateAttemptList.get(i).inputDoc.addError(Error.get("update",MongoCrudConstants.ERR_DOC_NO_LONGER_AVAILABLE,""));
+                        }
                     } catch (RuntimeException e) {
                     } finally {
                     }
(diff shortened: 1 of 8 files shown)


mongo/src/main/java/com/redhat/lightblue/mongo/crud/BasicDocSaver.java
```diff
@@ -358,10 +358,13 @@ private void updateDocs(CRUDOperationContext ctx,
...
                     try {
-                        Map<Integer,Error> errorMap=upd.commit();
-                        for(Map.Entry<Integer,Error> entry:errorMap.entrySet()) {
+                        BatchUpdate.CommitInfo ci=upd.commit();
+                        for(Map.Entry<Integer,Error> entry:ci.errors.entrySet()) {
                             updateAttemptList.get(entry.getKey()).inputDoc.addError(entry.getValue());
-                        }                        
+                        }
+                        for(Integer i:ci.lostDocs) {
+                            updateAttemptList.get(i).inputDoc.addError(Error.get("update",MongoCrudConstants.ERR_DOC_NO_LONGER_AVAILABLE,""));
```
alechenninger: Should these really be considered errors? We're just serializing operations that occurred concurrently. It's not an error that some no longer match the query, it's what should happen in order to establish the predictable happens-before relationship that ensures the atomicity we expect.

For a real world example, if I atomically find-and-update several documents only if they match a timestamp, and update that timestamp, two of these requests interweaving should produce the same results as two in serial. If they interweave and one fails, that's not the same result. If the concurrently changed docs that no longer match the timestamp are dropped, that does give us the same result. It's correct and expected way to recover from an error, not an error itself as I understand it.

bserdar:
This is a save operation, and at this point, we found that the document
being saved exists in the database so we decide to update it, but when we
try to update it, it fails. So I think this is an error at this point,
because that particular document was not saved.

alechenninger:
From what I can tell it looks like both save and update will consider any "lost docs" failures, when I'm not sure this is correct.

For updates, it's definitely not, since there is an explicit query. If a doc doesn't match the query, it's not a failure.

For save, what happens when you save without upsert a doc that isn't found? If it fails, then a lost doc should fail. If it doesn't fail, then a lost doc should not fail: it's the same scenario, a client doesn't care whether it interweaved with another request, that's just implementation detail. If the APIs were atomic at the DB level, then we'd have no interweaving at all. My point is that whether or not requests interleave and we have to serialize them, it shouldn't change the result compared to if they were already serialized. That's the whole point of serializing the operations I think.

alechenninger:
Similarly, if you do a save w/ upsert and the doc is modified such that it doesn't match the save any more during the tx, then a new doc should be inserted. This is the same as if this happened:

1. Thread 1 reads doc A
2. Thread 2 reads doc A
3. Thread 2 updates doc A to have id B. It's now doc B.
4. Thread 1 upserts doc A.

You will have two docs, doc A and doc B. Whether or not these threads' updates interleave within lightblue tx's at the same time doesn't matter; there is always an atomic moment at which a single thread starts the TX first and if that is thread 2, then it looks like above, regardless of whether it is first by a nanosecond or it is first by a day.

paterczm:
> For save, what happens when you save without upsert a doc that isn't found? If it fails, then a lost doc should fail. If it doesn't fail, then a lost doc should not fail.

I agree. I find the example above strange (involves changing identity, which is unusual and in case of mongo specific _id field not even possible), so here is another one:

1. Thread 1 reads doc A
2. Thread 2 removes doc A
3. Thread 1 updates doc A in memory

[ if upsert: false]
   4. Thread 1 persists to db. Since the doc has an _id, upsert: false and no doc with this _id already exists, nothing is updated. Client gets modified: 0. No error.
[ else if upsert: true ]
   4. Thread 1 persists to db. Since the doc has an _id, upsert: true and no doc with this _id already exists, it is inserted. Client gets modified: 1. No error.
[ /if ]

bserdar:
If upsert=false, and the doc doesn't exist, it is an error. We already give
an error for this case, ERR_SAVE_ERROR_INS_WITH_NO_UPSERT. I can return the
same error for this case,

I agree that if upsert=true, we should re-insert the doc. I'll do that.

paterczm:
> If upsert=false, and the doc doesn't exist, it is an error. We already give an error for this case, ERR_SAVE_ERROR_INS_WITH_NO_UPSERT

Mongo does not error in this case:
```
db.test.save({_id: 1, "foo": 2},{upsert: false})
WriteResult({ "nMatched" : 0, "nUpserted" : 0, "nModified" : 0 })
```

I wrongly assumed Lightblue behavior is the same. I agree we should keep existing behavior.

bserdar:
Pushed.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BatchUpdate.java
```diff
@@ -47,6 +49,17 @@
     public static final String DOCVER_FLD0=DocTranslator.HIDDEN_SUB_PATH.toString()+"."+DocVerUtil.DOCVER+".0";
 
 
+    public static class CommitInfo {
+        // errors[i] gives the error for the document i in the current batch
+        final public Map<Integer,Error> errors=new HashMap<>();
+        // An element of lostDocs is an index to a document in the current batch that is lost
+        final public Set<Integer> lostDocs=new HashSet<>();
+
+        public boolean hasErrors() {
```
alechenninger: I'd expect this not to look at lostDocs, since whether lostDocs are errors depends on context

mongo/src/main/java/com/redhat/lightblue/mongo/crud/IterateAndUpdate.java
```diff
@@ -223,15 +222,22 @@ public void update(CRUDOperationContext ctx,
...
-                                Map<Integer,Error> updateErrors=sup.commit();
+                                BatchUpdate.CommitInfo ci=sup.commit();
                                 measure.end("bulkUpdate");
-                                for(Map.Entry<Integer,Error> entry:updateErrors.entrySet()) {
+                                for(Map.Entry<Integer,Error> entry:ci.errors.entrySet()) {
                                     docUpdateAttempts.get(entry.getKey()+batchStartIndex).addError(entry.getValue());
                                 }
-                                int k=updateErrors.size();
+                                int k=ci.errors.size()+ci.lostDocs.size();
                                 numFailed+=k;
```
alechenninger: We agree updates should not consider lost docs errors right?

bserdar:
Changed that so numFailed only counts errors, numUpdated counts both errors
and lost docs.

bserdar:
That was added for testing, I removed it.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCrudConstants.java
```diff
@@ -51,6 +51,7 @@
     public static final String ERR_NO_SEQUENCE_NAME = "mongo-crud:NoSequenceName";
 
     public static final String ERR_CONCURRENT_UPDATE = "mongo-crud:ConcurrentUpdate";
+    public static final String ERR_DOC_NO_LONGER_AVAILABLE="mongo-crud:DocumentNoLongerAvailable";
```
alechenninger: This is no longer used now I think?

bserdar:
Removed.

mongo/src/main/java/com/redhat/lightblue/mongo/crud/BatchUpdate.java
```diff
@@ -47,6 +49,14 @@
     public static final String DOCVER_FLD0=DocTranslator.HIDDEN_SUB_PATH.toString()+"."+DocVerUtil.DOCVER+".0";
 
 
+    public static class CommitInfo {
+        // errors[i] gives the error for the document i in the current batch
+        final public Map<Integer,Error> errors=new HashMap<>();
+        // An element of lostDocs is an index to a document in the current batch that is lost
```
alechenninger: Probably worth making these javadoc

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoSafeUpdateProtocol.java
```diff
@@ -229,24 +229,24 @@ public int getSize() {
...
         while(nRetries-->0) {
             // Get the documents with concurrent modification errors
-            List<Integer> failedDocs=getFailedDocIndexes(results);
+            List<Integer> failedDocs=getFailedDocIndexes(ci);
             if(!failedDocs.isEmpty()) {
-                failedDocs=retryFailedDocs(failedDocs,results);
+                failedDocs=retryFailedDocs(failedDocs,ci);
             } else {
                 break;
             }
```
alechenninger: Might be clearer to do `if (failedDocs.isEmpty()) { break; }` without an else

PR #380, opened 2017-08-25
Update for new version of CRUDHealth

derek63:


mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -1122,47 +1121,47 @@ private Error analyzeException(Exception e, final String otherwise, final String
...
+            configDetails.put("maxResultSetSize", config.getMaxResultSetSize());
+            configDetails.put("maxQueryTimeMS", config.getMaxQueryTimeMS());
+            configDetails.put("mongoClient", config.getMongoClient());
+            configDetails.put("mongoClientOptions", config.getMongoClientOptions());
+            configDetails.put("readPreference", config.getReadPreference());
+            configDetails.put("server", config.getServer());
+            List<ServerAddress> serverAddresses = new ArrayList<>();
+            config.getServerAddresses().forEachRemaining(serverAddresses::add);
+            configDetails.put("serverAddresses", serverAddresses);
+            configDetails.put("writeConcern", config.getWriteConcern());
```
alechenninger: Awesome :-)

mongo/src/main/java/com/redhat/lightblue/mongo/crud/MongoCRUDController.java
```diff
@@ -1122,47 +1121,47 @@ private Error analyzeException(Exception e, final String otherwise, final String
...
                 } else {
-                    details.add(
-                            new StringBuilder(getMongoConfigDetails(config)).append("=>").append("ping:OK").toString());
+                    isHealthy = true;
                 }
             } catch (Exception e) {
                 isHealthy = false;
-                details.add(new StringBuilder(getMongoConfigDetails(config)).append("=>").append("ping_error:")
-                        .append(e.getMessage()).toString());
+                details.put("stackTrace", e);
```
alechenninger: Do you know if jackson will serialize the Exception instance nicely?

derek63: It comes out as follows:

`"stackTrace": "com.mongodb.MongoTimeoutException: Timed out after 30000 ms while waiting for a server that matches ReadPreferenceServerSelector{readPreference=primary}. Client view of cluster state is {type=STANDALONE, servers=[{address=localhost:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketOpenException: Exception opening socket}, caused by {java.net.ConnectException: Connection refused}}]",`

So it looks like the full stack trace is not being included, just the message and caused by.  I think that is probably sufficient here, but let me know if you disagree..

alechenninger: That looks good to me, though I'd probably change the field name from "stackTrace" to "exception" or something like that.

PR #385, opened 2017-09-25
Add health details at individual database level

svkm102:
The earlier implementation has a bug which can result in /diagnostics endpoint reporting MongoCRUDController. isHealthy : "true" even if connection fails for one of the database because the looping swallows the older isHealthy value and uses the latest isHealthy value intead.

To avoid such instances (however unlikely they may be), this PR includes a isHealthy attribute at each database level along with a overall isHealthy status for /diagnostics call.


svkm102:
Added test case

PR #404, opened 2025-06-06
Update mongo-driver-legacy version

derek63:


davidlanouette:
LGTM.  Although, this seems to conflict with the previous MR you created.

